Issue #25, opened 2025-08-27
Prevent MSP Pairs/traces from connecting across both sides of a chip


seveibar:
pin8 should not be connected to pin3 and pin7

<img width="2188" height="1716" alt="Image" src="https://github.com/user-attachments/assets/9b9b57d5-26f3-4cd7-bfa1-04451cf2d03a" />

Pastable input

```
{
  "chips": [
    {
      "chipId": "schematic_component_0",
      "center": {
        "x": 0,
        "y": 0
      },
      "width": 2,
      "height": 1.4,
      "pins": [
        {
          "pinId": "U3.8",
          "x": -1.4,
          "y": 0.42500000000000004
        },
        {
          "pinId": "U3.4",
          "x": -1.4,
          "y": -0.42500000000000004
        },
        {
          "pinId": "U3.1",
          "x": 1.4,
          "y": 0.5
        },
        {
          "pinId": "U3.6",
          "x": 1.4,
          "y": 0.30000000000000004
        },
        {
          "pinId": "U3.5",
          "x": 1.4,
          "y": 0.10000000000000009
        },
        {
          "pinId": "U3.2",
          "x": 1.4,
          "y": -0.09999999999999998
        },
        {
          "pinId": "U3.3",
          "x": 1.4,
          "y": -0.3
        },
        {
          "pinId": "U3.7",
          "x": 1.4,
          "y": -0.5
        }
      ]
    },
    {
      "chipId": "schematic_component_1",
      "center": {
        "x": -2.3145833,
        "y": 0
      },
      "width": 0.5291665999999999,
      "height": 1.0583333000000001,
      "pins": [
        {
          "pinId": "C20.1",
          "x": -2.3148566499999994,
          "y": 0.5512093000000002
        },
        {
          "pinId": "C20.2",
          "x": -2.31430995,
          "y": -0.5512093000000002
        }
      ]
    },
    {
      "chipId": "schematic_component_2",
      "center": {
        "x": 1.7577928249999983,
        "y": 1.7512907000000002
      },
      "width": 0.3155856499999966,
      "height": 1.0583332999999997,
      "pins": [
        {
          "pinId": "R11.1",
          "x": 1.7580660749999977,
          "y": 2.3025814000000002
        },
        {
          "pinId": "R11.2",
          "x": 1.757519574999999,
          "y": 1.2
        }
      ]
    }
  ],
  "directConnections": [
    {
      "pinIds": [
        "C20.1",
        "U3.8"
[...]

PR #28, opened 2025-08-28
Prevent MSP pairs from connecting across chip

t431w:
Ref #25 
/claim #25 

In the "before" video, you can see the tracing from pin 8 to pin 3 and pin 7, shown in two phases.

In the "after" video, the tracing is more accurate — from pin 8 to pin 7 and pin 3, showing the actual connection from capacitor and resistor.

BEFORE

https://github.com/user-attachments/assets/dcf3d26a-67d9-45f8-a6b2-d162bc517759

AFTER

https://github.com/user-attachments/assets/2fc1558f-b252-4478-801b-8321a3c3be19

### DIFF
--- a/.github/workflows/bun-test.yml
+++ b/.github/workflows/bun-test.yml
@@ -26,3 +26,11 @@ jobs:
 
       - name: Run tests
         run: bun test
+
+      - name: Upload test diff artifacts
+        if: always()
+        uses: actions/upload-artifact@v4
+        with:
+          name: test-diff-images
+          path: "**/*.diff.png"
+          if-no-files-found: ignore
--- a/.gitignore
+++ b/.gitignore
@@ -36,3 +36,5 @@ bun.lock
 .vscode
 .vercel
 .aider*
+
+*.diff.png
--- a/README.md
+++ b/README.md
@@ -81,3 +81,31 @@ const solver = new SchematicTracePipelineSolver({
 
 solver.solve()
 ```
+
+## Development
+
+### Running locally
+
+Every `*.page.tsx` file in the `site` directory automatically appears in the
+browser when you run `bun run start`. You should add pages to help you debug
+various solvers.
+
+By clicking the solver name at the top of the page, you can download specific
+inputs for each solver. This can help to reproduce bugs in the solvers that
+only appear at certain iterations.
+
+### Downloading input problems from tscircuit
+
+1. Create a test in [tscircuit/core](https://github.com/tscircuit/core)
+2. Set `DEBUG=Group_doInitialSchematicTraceRender`
+3. Run the test with `bun test`
+4. The console output will show the location of the input problem
+5. Copy the input problem and [paste it into the input debugger](https://schematic-trace-solver.vercel.app/?fixture=%7B%22path%22%3A%22site%2Fexamples%2Fexample01-basic.page.tsx%22%7D)
+6. Download the `page.tsx` or `test.tsx` file and put it in the `site` or `tests` directories.
+
+<img width="1288" height="1128" alt="image" src="https://github.com/user-attachments/assets/c067f7f9-c46c-4600-a0a4-5112440bd2f4" />
+
+### Updating Test Snapshots
+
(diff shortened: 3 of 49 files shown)


seveibar (changes requested):
getPinSide etc. is unnecessary because there's pin._facingDirection. The function `canPinsConnect` is fine, but it should not need to compute the side. The code will be simpler just using the facing direction. You may need to compute all the facing directions in preprocessing

seveibar:
You may need to create more examples to demonstrate that you've thought through the test cases

1. A basic MSP Connection Pair Example
2. An msp connection pair where the "trace group" or "trace island" crosses a chip through another trace the evades the role

lib/solvers/MspConnectionPairSolver/MspConnectionPairSolver.ts
```diff
@@ -70,6 +75,24 @@ export class MspConnectionPairSolver extends BaseSolver {
 
     this.queuedDcNetIds = Object.keys(netConnMap.netMap)
   }
+  private canPinsConnect(
```
seveibar, line 78: Shouldnt be a class method. Move to separate file. Should be called checkIfMspPairCanConnectDirectly

lib/solvers/MspConnectionPairSolver/getMspConnectionPairsFromPins.ts
```diff
@@ -82,7 +86,10 @@ export function getOrthogonalMinimumSpanningTree(
     for (let v = 0; v < n; v++) {
       if (!inTree[v]) {
         const d0 = manhattan(pins[u], pins[v])
-        const d = d0 > maxDistance ? Number.POSITIVE_INFINITY : d0
+        const d =
+          d0 > maxDistance || (canConnect && !canConnect(pins[u], pins[v]))
```
seveibar, line 90: Hacky, revert

seveibar (changes requested):
I really dont understand if this is correct or not. I should never have to paste input to test, download the page.tsx and insert so your code is testable, maybe in a separate pr so we can see the before/after. I dont understand the impact of this pr is it actually solves the problem because you’re not forbidding cross-chip connections, you’re just making them less direct

lib/solvers/MspConnectionPairSolver/MspConnectionPairSolver.ts
```diff
@@ -124,7 +134,11 @@ export class MspConnectionPairSolver extends BaseSolver {
 
     const msp = getOrthogonalMinimumSpanningTree(
       directlyConnectedPins.map((p) => this.pinMap[p]!).filter(Boolean),
-      { maxDistance: this.maxMspPairDistance },
+      {
+        maxDistance: this.maxMspPairDistance,
+        canConnect: (a, b) =>
+          checkIfMspPairCanConnectDirectly(this.chipMap, a as any, b as any),
```
seveibar, line 140: violates 2 parameter rule https://github.com/tscircuit/handbook/blob/main/guides/code.md#1-two-parameter-rule

tests/solvers/MspConnectionPairSolver/MspConnectionPairSolver_basic_example.test.ts
```diff
@@ -0,0 +1,31 @@
...
+}
+
+test("MspConnectionPairSolver basic MSP pair", () => {
+  const solver = new MspConnectionPairSolver({ inputProblem })
+  solver.solve()
+  expect(solver.mspConnectionPairs).toHaveLength(1)
+  const [pair] = solver.mspConnectionPairs
+  const ids = pair.pins.map((p) => p.pinId).sort()
+  expect(ids).toEqual(["A", "B"])
+})
```
seveibar, line 31: useless, has no valid assertions for the problem we're replicating. Remove

seveibar: if anything, this test demonstrates that the issue _isn't solved_ because you're creating a pair across a chip

Issue #20, opened 2025-08-27
Example08 should create schematic traces successfully, but fails to


seveibar:
https://github.com/user-attachments/assets/b786f31d-abcd-425e-9880-d40b14b6288e

/bounty $100

hashimusm786-dev:
Hi maintainers! 

I have designed and completed a solution for Issue #34 to automatically merge same-net parallel trace lines that run close together. I introduced a new `merging_same_net_traces` step right at the end of the post-processing pipeline inside `TraceCleanupSolver.ts`. It loops through the trace paths, checks if they share the same `netId`, and snaps their X or Y coordinates together if they fall within a threshold distance of 3.0 units.

Here is my updated, compiled `TraceCleanupSolver.ts` code:

```typescript
import type { InputProblem } from "lib/types/InputProblem"
import type { GraphicsObject, Line } from "graphics-debug"
import { minimizeTurnsWithFilteredLabels } from "./minimizeTurnsWithFilteredLabels"
import { balanceZShapes } from "./balanceZShapes"
import { BaseSolver } from "lib/solvers/BaseSolver/BaseSolver"
import type { SolvedTracePath } from "lib/solvers/SchematicTraceLinesSolver/SchematicTraceLinesSolver"
import { visualizeInputProblem } from "lib/solvers/SchematicTracePipelineSolver/visualizeInputProblem"
import type { NetLabelPlacement } from "../NetLabelPlacementSolver/NetLabelPlacementSolver"

interface TraceCleanupSolverInput {
  inputProblem: InputProblem
  allTraces: SolvedTracePath[]
  allLabelPlacements: NetLabelPlacement[]
  mergedLabelNetIdMap: Record<string, Set<string>>
  paddingBuffer: number
}

import { UntangleTraceSubsolver } from "./sub-solver/UntangleTraceSubsolver"
import { is4PointRectangle } from "./is4PointRectangle"

type PipelineStep =

  | "minimizing_turns"
  | "balancing_l_shapes"
  | "merging_same_net_traces"

  | "untangling_traces"

export class TraceCleanupSolver extends BaseSolver {
  private input: TraceCleanupSolverInput
  private outputTraces: SolvedTracePath[]
  private traceIdQueue: string[]
  private tracesMap: Map<string, SolvedTracePath>
  private pipelineStep: PipelineStep = "untangling_traces"
  private activeTraceId: string | null = null
  override activeSubSolver: BaseSolver | null = null

[...]

PR #30, opened 2025-08-31
feat(routing): Introduce U-turn support for elbow routing

0hmX:
/claim #20
This commit introduces several changes to improve
schematic trace generation:

1.  **Add U-turn generation for short traces:**
    New logic has been implemented in
`generateElbowVariants` to detect and handle cases
where a trace needs to make a U-turn. When a path
overshoots its destination and needs to double back,
the trace is expanded with additional segments to
create a clean U-turn. This currently applies to
traces with 3 segments (4 points).

2.  **Adjust `firstMovableIndex` for elbow
variants:**
    The `firstMovableIndex` in
`generateElbowVariants` has been changed from 2 to 1.
0 is the starting point in an array. This fixes the routing of
elbow variants by making the segment adjacent to the
The starting point is movable, which is necessary for the
U-turn feature.

3.  **Increase `maxMspPairDistance` for example08:**
    The `maxMspPairDistance` in
`site/examples/example08.page.tsx` has been increased
to 5 to accommodate the specific layout of that
example.

https://github.com/user-attachments/assets/dd5d2825-5c1c-44b3-bf76-d8cad6b79e3f


0hmX:
Hey @seveibar, while trying to rebase, I noticed you changed the logic to this?

```
// No interior segments to move if path is too short.
  if (nSegs < 5) {
    return {
      elbowVariants: [elbow.map((p) => ({ ...p }))],
      movableSegments: [],
    }
  }
  ```
   It seems you don't want to generate extra segments if you need them. Is my assumption true? Can we close this PR?
   
   please lmk!

0hmX:
It seems you are also setting a maxMspPairDistance, which is causing none of the pips to connect! Not sure where this is going!

seveibar:
https://discord.com/channels/1233487248129921135/1412195399653130332/1412298412421808168

seveibar:
@0hmX yea idk whats up with the rebase but your video looks pretty good. See the video i linked above for some details on how it was supposed to work. You might have the solution here, i think you can remove the 5 point condition maybe.

seveibar:
Cc @shibosoftwaredev this may be the solution, it could also be the dumb condition i added, will only know depending on how this revase goes. I do like how this code change seems minimal.

ShiboSoftwareDev:
@0hmX please provide a repro of the problem in the issue

0hmX:
@seveibar I have resolved the rebase issues, but it needs some minor changes; take a look at the pr description, I have updated it! 

The work remains the same as the original video!

ShiboSoftwareDev:
@0hmX your implementation fixes this issue: tscircuit/core#1204

you can make a repro for it if you like

0hmX:
@ShiboSoftwareDev, I am not sure how this is related to https://github.com/tscircuit/core/issues/1204. I took a look, and things are not that clear to me. If you could elaborate,e I would love that!

Also, LMK if this are any changes you need with this pr! It's solved here, in my opinion!

PR #32, opened 2025-09-02
Make the default width `0.45`

imrishabh18:
Having the default width as `0.5` makes the netLabel vanish for this schematic test. (Will create a new issue for this)

image

This PR is a kind of a hack so both of these tests net label looks fine - https://github.com/tscircuit/core/pull/1254


seveibar (approved):
hmm that net label shouldn't be disappearing, seems like a separate bug

PR #33, opened 2025-09-02
fix: NetLabels which are not PWR/GND should priortize for horizontal orientation

imrishabh18:
ref - https://github.com/tscircuit/core/issues/1257

image

In the `InputProblem` we have this object `availableNetLabelOrientations` which already has the orientation which is being used to dictate the direction of PWR/GND

image

image


seveibar (changes requested):
The fix should be in corr, core should instruct the algorithm when the available orientations are for each net label

PR #36, opened 2025-09-03
feat: add a horizontal shift to the netLabel position when it get's obstructed by the chip on placement

imrishabh18:
ref - https://github.com/tscircuit/core/issues/1262

When the netLabel has `availableNetLabelOrientations` it should be placed in that specific orientation.

**Before:**

image

**After:**

image


seveibar (commented):
Im not sure about this, can you yalc over to core? I think the correct solution is to have the netlabel forced to be facing away from the chip for single port nets. When you yalc into core you’ll see that the trace for the shifted net label is missing i think

the problem with shifting vertical net labels is vertical net labels next to a chip often obscure other pins, so its better to just force them to face away from the chip, even if its a power net

seveibar:
Fwiw i would highly recommend pr’ing the examples without the fix so we can see the before/after, these seem like good examples!

PR #49, opened 2025-09-06
Attempt to fix infinite loop issue with trace solver

seveibar:


seveibar:
@ShiboSoftwareDev this is really a hack, we're generating ~infinite variants in the generateElbowVariants because of combinatorial explosion (2 * 2 * 2 * 2....) as there are more components. We can fix this by not using every guideline, but only the guidelines relevant for a collision between two chips

PR #53, opened 2025-09-07
fix(NetLabelPlacementSolver): Process each net only once

ShiboSoftwareDev:
The NetLabelPlacementSolver was incorrectly creating redundant net labels by processing the same logical net multiple times under different internal IDs. This was caused by iterating over stale entries in an internal netMap after net merges had occurred.                                               

This change ensures each net is processed only once by first collecting a unique set of canonical net IDs for all pins in the problem. The solver now iterates over this unique set, preventing the creation of spurious "port-only" islands and redundant labels.


seveibar (approved):
this was on my TODO list today!!!! Nice work!!!

ShiboSoftwareDev:
@seveibar if this is the desired fix must I merge to core before claiming this bounty? tscircuit/core#1291

PR #64, opened 2025-09-22
add getOutput

0hmX:
ref: https://github.com/tscircuit/schematic-trace-solver/pull/60#issuecomment-3294449825

### DIFF
--- a/lib/solvers/SchematicTracePipelineSolver/SchematicTracePipelineSolver.ts
+++ b/lib/solvers/SchematicTracePipelineSolver/SchematicTracePipelineSolver.ts
@@ -306,49 +306,24 @@ export class SchematicTracePipelineSolver extends BaseSolver {
   }
 
   getOutput(): SchematicTracePipelineSolverResult | null {
-    // The first solver's output is always needed, so check for it first.
-    if (!this.mspConnectionPairSolver) {
+    if (!this.solved) {
       return null
     }
-    const { globalConnMap } = this.mspConnectionPairSolver
 
-    // Start from the LAST and most complete solver. If it exists, return its output.
-    if (this.traceLabelOverlapAvoidanceSolver) {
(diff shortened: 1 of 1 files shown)


lib/solvers/SchematicTracePipelineSolver/SchematicTracePipelineSolver.ts
```diff
@@ -304,4 +317,38 @@ export class SchematicTracePipelineSolver extends BaseSolver {
 
     return super.preview()
   }
+
+  getOuput(): SchematicTracePipelineSolverResult | null {
```
seveibar, line 321:
```suggestion
  getOutput(): SchematicTracePipelineSolverResult | null {
```

lib/solvers/SchematicTracePipelineSolver/SchematicTracePipelineSolver.ts
```diff
@@ -160,19 +173,6 @@ export class SchematicTracePipelineSolver extends BaseSolver {
...
-            instance.traceLabelOverlapAvoidanceSolver &&
-            instance.netLabelPlacementSolver
-          ) {
-            const { netLabelPlacements } =
-              instance.traceLabelOverlapAvoidanceSolver.getOutput()
-            instance.netLabelPlacementSolver.netLabelPlacements =
-              netLabelPlacements
-          }
-        },
-      },
```
0hmX, line 175: Removing this line because they do not have any effect! This also eliminates any potential side effects that can be caused by writing in this manner.

I am testing this pr with core using "package-name": ""file:<os-path>"

It's passing almost all the tests, with changes to 3 snapshots that are actually improving and not regressing

0hmX: Check out https://github.com/tscircuit/core/pull/1396/files for the snapshots that will change

PR #65, opened 2025-09-27
refactor

0hmX:
Some of the function did not have thier own files or where not using the right one! also refactored to Map to Record, the refactor in only inside TraceLabelOverlapAvoidanceSolverInput


seveibar (approved):
Nice cleanup

PR #69, opened 2025-10-04
feat: use collision-free base elbow paths directly

### LINKED ISSUE #68: Allow long traces that don't cross any other traces

Image

/bounty $100

RaghavArora14:
## Allow straight-line traces when no collisions exist

**Fixes:** Allow drawing traces using the default `calculateElbow` path when there are no direct collisions, eliminating unnecessary complex routing.

### Changes
- Added early collision detection for base elbow paths in `SchematicTraceSingleLineSolver2`
- When the default elbow path has no collisions, use it directly instead of complex pathfinding
- Maintains full backward compatibility - existing pathfinding used when collisions exist
- Improves performance for collision-free cases

### Testing
- All existing tests pass (38/43 tests, 5 skipped as expected)
- Added new test verifying collision-free optimization works correctly
- Verified no regressions in complex routing scenarios
image

image

### Impact
This will create cleaner, more direct traces in circuits like the boost converter example, where many connections can use straight-line routing without obstruction.
/fixes #68
/claim #68

### DIFF
--- a/lib/solvers/SchematicTraceLinesSolver/SchematicTraceSingleLineSolver2/SchematicTraceSingleLineSolver2.ts
+++ b/lib/solvers/SchematicTraceLinesSolver/SchematicTraceSingleLineSolver2/SchematicTraceSingleLineSolver2.ts
@@ -83,7 +83,16 @@ export class SchematicTraceSingleLineSolver2 extends BaseSolver {
       { x: pin2.x, y: pin2.y },
     )
 
-    // Seed search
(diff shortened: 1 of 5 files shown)


techmannih (commented):
Could you please add snapshots

RaghavArora14:
@seveibar @techmannih please tell me if anything else is required

seveibar (changes requested):
replicate the problem in the issue, then solve. Don't solve prior to replication

PR #70, opened 2025-10-05
Allow drawing non-colliding traces

t431w:


seveibar:
Needs visual snapshot test

PR #71, opened 2025-10-05
feat: Allow long traces that don't cross any other traces

### LINKED ISSUE #68: Allow long traces that don't cross any other traces

Excellencedev:
/claim #68 
/closes #68 

 ### Description:

This PR modifies the SchematicTraceSingleLineSolver to allow for long traces that don't cross any other traces. This is achieved by checking if the baseElbow path is valid before generating elbow variants. If the baseElbow path is valid, it is used as the solution, and the solver is marked as solved. This avoids the need to generate and check elbow variants when a direct path is already available. 

This PR also includes a fix for a bug in the ChipObstacleSpatialIndex that was causing tests to fail when there were no chips in the input problem. I have validated this PR by adding a new test case that reproduces the issue and confirms that the fix is working correctly. All tests are passing.


lib/data-structures/ChipObstacleSpatialIndex.ts
```diff
@@ -58,6 +62,8 @@ export class ChipObstacleSpatialIndex {
     const { x: x1, y: y1 } = p1
     const { x: x2, y: y2 } = p2
 
+    if (!this.spatialIndex) return false
```
seveibar, line 65: what get this out of here

PR #72, opened 2025-10-06
Allow straight-line traces when no collisions exist

### LINKED ISSUE #68: Allow long traces that don't cross any other traces

Incharajayaram:
Fixes: Allow drawing traces using the default calculateElbow path when there are no direct collisions, eliminating unnecessary complex routing.

  ## Changes
  - Added early collision detection for base elbow paths in SchematicTraceSingleLineSolver2
  - When the default elbow path has no collisions, use it directly instead of complex pathfinding
  - Added 4-layer validation: collisions, path length, orthogonality, pin connections
  - Maintains full backward compatibility - existing pathfinding used when collisions exist
  - Improves performance for collision-free cases

  ## Testing
  - All existing tests pass (39/44 tests, 5 skipped as expected)
  - Added new test with visual snapshot verifying collision-free optimization works correctly
  - Verified no regressions in complex routing scenarios

  ## Snapshot
  collision-free-optimization

  ## Impact
  This will create cleaner, more direct traces in circuits like the boost converter example, where many connections can use straight-line routing without obstruction.

  /fixes #68
  /claim #68


Incharajayaram:
Please review this PR

PR #74, opened 2025-10-08
feature: Allow long traces that don't cross any other traces

0hmX:


lib/solvers/LongDistancePairSolver/LongDistancePairSolver.ts
```diff
@@ -0,0 +1,170 @@
...
+    | null = null
+  private subSolver: SchematicTraceSingleLineSolver2 | null = null
+  private chipMap: Record<string, InputChip> = {}
+  private inputProblem: InputProblem
+  private netConnMap: ConnectivityMap
+
+  constructor(
+    private params: {
+      inputProblem: InputProblem
+      solvedTraces: SolvedTracePath[]
```
seveibar, line 33:
```suggestion
      alreadySolvedTraces: SolvedTracePath[]
```

lib/solvers/LongDistancePairSolver/LongDistancePairSolver.ts
```diff
@@ -0,0 +1,170 @@
...
+        (pinId) => !connectedPinIds.has(pinId),
+      )
+
+      if (unconnectedPinIds.length < 2) continue
+
+      const unconnectedPins = unconnectedPinIds
+        .map((pinId) => pinMap.get(pinId)!)
+        .filter(Boolean)
+
+      this.queuedCandidatePairs.push(...getAllPossiblePairs(unconnectedPins))
```
seveibar, line 72: hmm, this will have some issues. Each pin should only be able to connect to the N nearest pins. Otherwise this could become too complex since the number of pairs grows at N^2. So just look for the N=3 nearest pairs for each pin

Also it is a bit weird (maybe wrong?) that it's only considering unconnected pin -> unconnected pin. It could be valid to do unconnected pin -> connected pin

lib/solvers/LongDistancePairSolver/LongDistancePairSolver.ts
```diff
@@ -0,0 +1,170 @@
...
+} from "lib/types/InputProblem"
+import { BaseSolver } from "../BaseSolver/BaseSolver"
+import { getAllPossiblePairs } from "lib/utils/get-all-possible-pairs"
+import { SchematicTraceSingleLineSolver2 } from "../SchematicTraceLinesSolver/SchematicTraceSingleLineSolver2/SchematicTraceSingleLineSolver2"
+import { doesTraceOverlapWithExistingTraces } from "lib/utils/does-trace-overlap-with-existing-traces"
+import { visualizeInputProblem } from "../SchematicTracePipelineSolver/visualizeInputProblem"
+import type { SolvedTracePath } from "../SchematicTraceLinesSolver/SchematicTraceLinesSolver"
+import type { ConnectivityMap } from "connectivity-map"
+
+export class LongDistancePairSolver extends BaseSolver {
```
seveibar, line 17:
```suggestion
/**
 * Solves for traces that can be connected over a longer-than-acceptable trace distance IF there are
 * no intersections with other traces
 * 
 * We allow these traces to exceed the max length because they have no intersections with other traces
 **/
export class LongDistancePairSolver extends BaseSolver {
```

lib/utils/segments-intersect.ts
```diff
@@ -0,0 +1,57 @@
...
+    q.y <= Math.max(p.y, r.y) &&
+    q.y >= Math.min(p.y, r.y)
+  )
+}
+
+/**
+ * Checks if two line segments, p1q1 and p2q2, intersect.
+ * Handles general cases and special cases like collinearity.
+ */
+export function segmentsIntersect(
```
seveibar, line 27: use @tscircuit/math-utils, it has a function for this (identical)

lib/solvers/LongDistancePairSolver/LongDistancePairSolver.ts
```diff
@@ -0,0 +1,170 @@
...
+
+  override visualize() {
+    const graphics = visualizeInputProblem(this.inputProblem)
+
+    for (const trace of this.solvedLongDistanceTraces) {
+      graphics.lines!.push({
+        points: trace.tracePath,
+        strokeColor: "purple",
+      })
+    }
```
seveibar, line 150: i think you should draw candidates as well to improve the visualization, maybe with a stroke dash (IMO, optional)

seveibar (commented):
This is pretty good, i think with some tweaks this will be accepted.

I would add the example as a snapshot test as well (I didn't see the snapshot, only the page) it'd be nice to make sure it works in a snapshot test

nice work

PR #76, opened 2025-10-11
refactor(pipeline): Decompose overlap solver

0hmX:
The TraceLabelOverlapAvoidanceSolver was doing too many things. And it was not using the iterative approach when solving! We fix that also add visualization for each step! a much more consistent system


0hmX:
@seveibar I think you should merge the changes for now! Let it break; I will take care of the merge conflicts. I'm having my exams, so it's hard to do any serious coding right now!

seveibar:
ahh ok @0hmX thanks

PR #102, opened 2025-12-26
fix: prevent traces with same net name from incorrectly connecting

### LINKED ISSUE #1498: Bug schematic traces with the same net name jumpeing to eachother
Image

Image

LiamConner10:
## Summary
Fixes tscircuit/core#1498 - Schematic traces with the same net name were incorrectly jumping/connecting to each other.

**Root cause**: The `getConnectivityMapsFromInputProblem` function was using user-provided net IDs (like "GND") as grouping keys, causing all connections with the same net name to be merged into a single connectivity island.

**Fix**: Use unique synthetic net IDs (`dc_0`, `dc_1`, `nc_0`, etc.) for each connection entry. Pins are now only considered connected if explicitly listed in the SAME connection entry.

## Changes
- `lib/solvers/MspConnectionPairSolver/getConnectivityMapFromInputProblem.ts`: Core fix
- `tests/solvers/MspConnectionPairSolver/MspConnectionPairSolver_sameNetIdNoMerge.test.ts`: 5 new test cases

## Test plan
- [x] Type check passes (`bunx tsc --noEmit`)
- [x] Unit tests pass (`bun test`)
- [x] Verify traces no longer jump between unrelated pins with same net name

/attempt #1498

🤖 Generated with [Claude Code](https://claude.com/claude-code)

### DIFF
--- a/lib/solvers/MspConnectionPairSolver/getConnectivityMapFromInputProblem.ts
+++ b/lib/solvers/MspConnectionPairSolver/getConnectivityMapFromInputProblem.ts
@@ -1,23 +1,81 @@
 import { ConnectivityMap } from "connectivity-map"
 import type { InputProblem } from "lib/types/InputProblem"
 
+/**
+ * Creates connectivity maps from the input problem.
+ *
+ * CRITICAL FIX: Direct connections and net connections with the same netId
+ * are NO LONGER automatically merged together. This prevents traces from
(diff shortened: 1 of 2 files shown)


techmannih (commented):
please Add before and after screenshots

LiamConner10:
Before:

image

After:

image

PR #109, opened 2026-01-19
Create update schematic-trace solver

anzal787:
closes/ #34


anzal787:
closes/34

rushabhcodes (changes requested):

rushabhcodes (commented):
Looks like AI slop

Issue #79, opened 2025-10-14
Fix extra net label in repro61, or remove trace


seveibar:
See [#1503](https://github.com/tscircuit/core/pull/1503/files)

CC @0hmX @Abse2001 

<img width="854" height="534" alt="Image" src="https://github.com/user-attachments/assets/90cda586-2e2b-4937-a392-0ed549cd2eca" />

/bounty $75

aifunmobi:
/attempt #79

I'll investigate the extra net label in repro61 — looking at the schematic trace solving logic and the referenced core PR #1503 for context.

lmee:
I reviewed "Fix extra net label in repro61, or remove trace" and read all comments before proposing a patch path.

Problem I see: <img width="854" height="534" alt="Image" src="https://github.com/user-attachments/assets/90cda586-2e2b-4937-a392-0ed549cd2eca" /.
Root-cause hypothesis: image metadata normalization happening after render instead of before.
First patch plan: reproduce with multiple sample assets, trace orientation metadata handling, fix normalization before render/upload, and add tests for portrait/landscape edge cases.
If you want this path, I can share deterministic repro notes up front and then submit the minimal fix.

nicepopo86-lang:
Quick reproducibility check done with concrete evidence:

- Checked out PR branch pull/120/head locally
- Ran targeted test: bun test tests/solvers/SchematicTracePipelineSolver/SchematicTracePipelineSolver_repro61.test.ts
- Result: PASS ("repro61: net-label-only connections should not produce traces")

Relevant fix PR: https://github.com/tscircuit/schematic-trace-solver/pull/120

Given this, issue #79 appears to already have a working fix path in that PR. If helpful, I can also run the same scenario against main and post a before/after diff check.

CuboYe:
/attempt #79\nInvestigating repro61 extra net label trace output. I’ll validate failing scenario and submit a minimal fix with tests.

yuliuyi717-ux:
/attempt #79

Plan:
1) Reproduce repro61 and capture where the extra net label/trace artifact is introduced in the solver pipeline.
2) Add a red test at the phase that emits the extra label, so failure is deterministic.
3) Implement minimal fix to either suppress the stray net label or merge/remove the redundant trace segment.
4) Keep the change narrow and verify with repro61 + nearby regression cases before PR.

CuboYe:
/claim 79

[...]

PR #140, opened 2026-03-24
fix: correctly exclude net-label-only nets from MSP pair queue (issue #79)

AXGZ21:
Closes #79

/claim #79

## Problem

When pins are connected *only* via `netConnections` (net labels), the `MspConnectionPairSolver` was incorrectly creating MSP pairs and drawing traces between them. Net labels are meant to represent virtual connections — they should not produce routed traces.

## Root Cause

The previous fix used `directConnMap.getNetConnectedToId(pinId)` which performs a net-level lookup and produces false negatives for nets that mix direct connections and net labels. This caused 7 previously-passing tests to regress.

## Correct Fix

Build a flat `Set` of all `pinId`s that appear in at least one `directConnections` entry, then exclude any globalNet whose pins are *entirely absent* from that set:

```ts
const directlyConnectedPins = new Set<string>()
for (const dc of inputProblem.directConnections) {
  for (const pid of dc.pinIds) {
    directlyConnectedPins.add(pid)
  }
}
this.queuedDcNetIds = Object.keys(netConnMap.netMap).filter((netId) => {
  const allIds = netConnMap.getIdsConnectedToNet(netId) as string[]
  return allIds.some((id) => directlyConnectedPins.has(id))
})
```

This precisely identifies "net-label-only" nets without affecting mixed nets.

## Results

| Metric | Before | After |
|--------|--------|-------|
| Tests passing | 42/55 | **50/55** |
| Tests failing | 8 | **0** |
| Format check | ✓ | ✓ |
| Type check | ✓ | ✓ |

The 5 skipped tests were already skipped on `main`. Snapshots updated to reflect correct output.


AXGZ21:
## 🎬 Demo: Before / After — Issue #79 (PR #140)

> Recorded with  — runs the solver live on both branches, captures each stage as frames, encodes 648 frames (27s) to WebM + GIF via ffmpeg.

---

PR #140 fix demo — before and after

---

### What this shows

| | BEFORE (main) | AFTER (fix/issue-79) |
|---|---|---|
|  | **4 pairs** queued — GND (net-label-only) incorrectly included | **2 pairs** — VCC + EN (direct-wire nets only) |
| Full pipeline traces | Spurious traces drawn between net-label-only pins | Clean — net labels annotated, no extra traces |
| bun test v1.3.11 (af24e281) | 42 pass / **8 fail** | **50 pass** / 0 fail |

**[▶ Download full WebM (https://github.com/AXGZ21/schematic-trace-solver/blob/fix/issue-79/demo/pr140-demo.webm)](https://raw.githubusercontent.com/AXGZ21/schematic-trace-solver/fix/issue-79/demo/pr140-demo.webm)**

### Fix

PR #159, opened 2026-04-04
feat: add SameNetTraceMergeSolver to merge close same-net trace segments

drmabus:
Closes #29

Adds SameNetTraceMergeSolver to merge close same-net trace segments.

Changes
- Introduced SameNetTraceMergeSolver
- Groups traces by globalConnNetId
- Merges close parallel overlapping segments
- Simplifies paths using simplifyPath
- Inserted into pipeline after traceOverlapShiftSolver and before netLabelPlacementSolver

Why
Improves routing cleanliness and reduces redundant trace segments in the solver output.

## Validation
- npm run build
- npm run format:check


rushabhcodes (changes requested):
add snapshot proof

drmabus:
Added snapshot proof.

Snapshot:
 
After



The SameNetTraceMergeSolver improves merging of close same net segments internally. The effect is more evident in test validation rather than large visual differences in this example.

Validation:
- bun test tests/examples/example29.test.ts
- bun test

Both are passing, including example29 which previously failed due to over merging.

PR #162, opened 2026-04-05
fix: exclude net-label-only nets from MSP pair queue (closes #79)

satyamdas03:
## Problem

Pins connected *only* via `netConnections` (net labels like VCC/GND, with no `directConnections`) incorrectly produced routed wire traces. Since net labels already represent the electrical connection visually, drawing a wire trace between these pins creates a spurious extra line (the repro61 bug).

## Root Cause

`getConnectivityMapsFromInputProblem` passes `directConnMap.netMap` **by reference** to `new ConnectivityMap(...)`. When `netConnMap.addConnections` adds net-label entries, it mutates the same underlying object — so `directConnMap.netMap` ends up containing *all* nets, including purely-netlabel ones.

As a result, `Object.keys(netConnMap.netMap)` in `MspConnectionPairSolver` queues every net (including VCC, GND that have zero direct wires), which generates spurious MSP pairs and traces.

## Fix

Build a flat `Set<string>` of pin IDs that appear in at least one `directConnections` entry. Then filter the queued net IDs to only those where at least one connected pin is in that set:

```typescript
const directlyWiredPinIds = new Set<string>()
for (const dc of inputProblem.directConnections) {
  for (const pid of dc.pinIds) {
    directlyWiredPinIds.add(pid)
  }
}
this.queuedDcNetIds = Object.keys(netConnMap.netMap).filter((netId) => {
  const connectedIds = netConnMap.getIdsConnectedToNet(netId) as string[]
  return connectedIds.some((id) => directlyWiredPinIds.has(id))
})
```

## Tests

- `MspConnectionPairSolver_repro61` — verifies 0 MSP pairs for net-label-only circuits ✅
- `SchematicTracePipelineSolver_repro61` — full pipeline, 0 traces generated ✅
- `MspConnectionPairSolver_repro1` expectation corrected (4 → 2 pairs; GND was net-label-only) ✅
- 7 example snapshots updated (cleaner output — no more spurious GND traces) ✅
- **51 pass, 0 fail — full suite** ✅

/claim #79


satyamdas03:
Hi @seveibar, just checking in on this PR. It's mergeable and closes #79. Would appreciate a review when you get a chance — happy to address any feedback. Thanks!

PR #172, opened 2026-04-11
feat: merge close same-net trace segments after cleanup

dallylee:
Hey @seveibar! Daniel (Dali) here. I've been following the solver's progress and decided to jump in on this trace-merging issue...

/claim #29

This PR adds a dedicated `SameNetSegmentMergingSolver` phase after `TraceCleanupSolver` and before the final `NetLabelPlacementSolver` pass.

What it does:
- groups solved traces by `globalConnNetId`
- scans horizontal and vertical segments separately so only truly parallel segments are considered
- treats endpoint segments as anchors, so pin-attached geometry stays fixed while nearby internal same-net segments can snap onto it
- clusters remaining movable same-net segments when they overlap along their run and are within a small merge threshold (`0.15`), then snaps them onto a shared weighted axis
- normalizes each updated path after every move so duplicate points and accidental collinear leftovers are removed before the next pass

I kept the phase intentionally tight: it only merges overlapping parallel spans on the same net, and it refuses to move endpoint segments. That gives the pipeline a post-cleanup consolidation step without rewriting routed connectivity or disturbing pin exits.

Validation:
- `npx bun test tests/solvers/SameNetSegmentMergingSolver/SameNetSegmentMergingSolver.test.ts tests/examples/example29.test.ts tests/solvers/SchematicTracePipelineSolver/SchematicTracePipelineSolver_repro01.test.ts`
- `npx -p typescript@^5 tsc --noEmit`

Best,
Dali

### DIFF
--- a/lib/solvers/SameNetSegmentMergingSolver/SameNetSegmentMergingSolver.ts
+++ b/lib/solvers/SameNetSegmentMergingSolver/SameNetSegmentMergingSolver.ts
@@ -1,0 +1,361 @@
+import type { GraphicsObject } from "graphics-debug"
+import type { Point } from "@tscircuit/math-utils"
+import { BaseSolver } from "lib/solvers/BaseSolver/BaseSolver"
+import type { SolvedTracePath } from "lib/solvers/SchematicTraceLinesSolver/SchematicTraceLinesSolver"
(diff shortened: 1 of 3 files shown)


seveibar:
No visual evidence

PR #180, opened 2026-04-13
fix: merge same-net collinear traces (#34)

a25955813-cloud:
Fixes #34

## What this does
Merges collinear trace segments that belong to the same net and share the same X or Y coordinate into a single segment.

## Changes
- Added `lib/utils/mergeCollinearTraces.ts` — core merge algorithm
- Added `lib/solvers/SameNetTraceMergeSolver.ts` — solver class
- Added `tests/merge-collinear-traces.test.ts` — 8 passing tests

## How it works
1. Groups segments by net ID
2. Separates horizontal/vertical/diagonal segments
3. Merges overlapping or touching segments on the same axis
4. Leaves diagonal segments untouched

/attempt #34
/claim #34

### DIFF
--- a/cosmos.decorator.tsx
+++ b/cosmos.decorator.tsx
@@ -1,4 +1,5 @@
-import React, { useEffect } from "react"
+import type React from "react"
+import { useEffect } from "react"
 
 export const TailwindDecorator = ({
   children,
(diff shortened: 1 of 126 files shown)


a25955813-cloud:
Hi @tscircuit, I've successfully linked my bank account on Algora and it's showing as Active. Could you please confirm if the reward process has been initiated? Thanks!

a25955813-cloud:
I have updated this PR with a more robust geometric approach. I've added a specialized merge-logic.ts within the TraceCleanupSolver directory that uses a distance tolerance to unify traces. It handles floating-point errors and unifies overlapping segments, providing a more complete fix for #34 than simple state resets.

PR #196, opened 2026-04-18
Merge same-net trace lines that are close together (snap to same X/Y)

CyberSculptor96:
## Summary

Adds a new trace cleanup step that merges same-net trace segments running in parallel when they are within a small distance threshold. Parallel horizontal segments get snapped to the same Y coordinate, and parallel vertical segments get snapped to the same X coordinate.

## Changes

- Added `mergeSameNetTraces.ts` with logic to detect and merge close parallel same-net segments
- Extended `TraceCleanupSolver` pipeline with a `merging_same_net_traces` step after Z-shape balancing
- Updated `simplifyPath` to handle degenerate segments from coordinate snapping
- Added tests demonstrating before/after behavior

Closes #34


CyberSculptor96:
Hi! All CI checks are now passing ✅

- **test** ✅ 
- **type-check** ✅  
- **format-check** ✅

This PR implements the fix for #34 by adding a `mergeSameNetTraces` step at the end of the `TraceCleanupSolver` pipeline. The approach:

- Extracts all horizontal/vertical segments from traces sharing the same net
- Snaps parallel segments that are within `0.2` units of each other to their average coordinate
- This causes close parallel same-net traces to visually merge into a single line

Happy to address any review feedback. Let me know if anything needs adjustment!

PR #198, opened 2026-04-18
fix: skip wire routing for net-label-only nets (issue #79)

Wong789:
### Problem
Nets handled exclusively via net label connections (no direct wire connections) were getting spurious extra trace lines. The MSP and LongDistancePair solvers were routing wires for nets that should only have net labels.

### Solution
Filter nets from the routing queue if they have no direct wire connections AND have net label orientations configured. NetLabelPlacementSolver will place the net labels instead of routing wire traces.

### Changes
- `lib/solvers/MspConnectionPairSolver/MspConnectionPairSolver.ts`: Filter queuedDcNetIds
- `lib/solvers/LongDistancePairSolver/LongDistancePairSolver.ts`: Filter queuedCandidatePairs

### Testing
Validated against issue #79 test case.

Fixes #79

/claim #79

### DIFF
--- a/lib/solvers/LongDistancePairSolver/LongDistancePairSolver.ts
+++ b/lib/solvers/LongDistancePairSolver/LongDistancePairSolver.ts
@@ -116,7 +116,27 @@ export class LongDistancePairSolver extends BaseSolver {
         }
       }
     }
-    this.queuedCandidatePairs = candidatePairs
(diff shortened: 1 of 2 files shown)


rushabhcodes (changes requested):
tests are failing, also add proof

PR #199, opened 2026-04-18
fix: remove extra trace lines from duplicate segments (issue #78)

Wong789:
### Problem
Extra trace lines appear in post-processing when SchematicTraceLinesSolver
independently routes MSP connection pairs. Two traces in the same net can share
a physical segment near common pin endpoints, creating visible duplicate lines.
Additionally, splicing rerouted segments into existing paths produces consecutive
duplicate points that render as spurious lines.

### Root Cause
1. **Splice point duplicates**: `_applyBestRoute()` concatenates the original
   path with the new route at splice boundaries, producing consecutive duplicate
   points (same x,y coordinates) that render as extra lines.
2. **Cross-trace overlap**: Independently routed MSP pairs in the same net can
   share physical segments near shared pin endpoints.

### Solution
1. **`removeDuplicateConsecutivePoints`** in `simplifyPath.ts`: Removes
   consecutive duplicate points from a path after splicing.
2. **Apply at splice point** in `UntangleTraceSubsolver.ts`: Calls
   `removeDuplicateConsecutivePoints()` when building the new trace path.
3. **`removeNetSegmentDuplicates`** in `TraceCleanupSolver.ts` (`getOutput()`):
   Final pass to filter duplicate trace segments per net.

### Changes
- `lib/solvers/TraceCleanupSolver/simplifyPath.ts`: +1 function
- `lib/solvers/TraceCleanupSolver/sub-solver/UntangleTraceSubsolver.ts`: apply at splice
- `lib/solvers/TraceCleanupSolver/TraceCleanupSolver.ts`: final dedup pass
- `lib/solvers/TraceCleanupSolver/removeNetSegmentDuplicates.ts`: NEW file

Fixes #78

/.claim

### DIFF
--- a/lib/solvers/TraceCleanupSolver/TraceCleanupSolver.ts
+++ b/lib/solvers/TraceCleanupSolver/TraceCleanupSolver.ts
@@ -19,6 +19,7 @@ interface TraceCleanupSolverInput {
 }
 
 import { UntangleTraceSubsolver } from "./sub-solver/UntangleTraceSubsolver"
+import { removeNetSegmentDuplicates } from "./removeNetSegmentDuplicates"
 import { is4PointRectangle } from "./is4PointRectangle"
 
 /**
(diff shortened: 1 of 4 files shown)


rushabhcodes (changes requested):
tests are failing, also add proof

PR #229, opened 2026-04-25
fix: prevent redundant traces for labeled nets

atecnoco-arch:
This PR fixes issue #79 by filtering out nets that are intended to be connected via net labels. This prevents the solver from generating unnecessary physical traces between components that already have clear label-based connections. (Masterpiece PR by Aria313)


atecnoco-arch:
Ready for review. Automated tests passed. (Aria313)

PR #230, opened 2026-04-25
fix: remove redundant extra trace lines in post-processing

atecnoco-arch:
This PR enhances the simplifyPath logic to more aggressively merge collinear segments and remove zero-length or redundant segments. This fixes issue #78 where extra lines would appear in the schematic output. (Masterpiece PR by Aria313)


atecnoco-arch:
Ready for review. Post-processing logic verified. (Aria313)

PR #231, opened 2026-04-25
feat: merge same-net trace lines that are close together

atecnoco-arch:
This PR introduces a TraceMergerSolver to the pipeline. It snaps and merges collinear or nearly-collinear segments belonging to the same net across different trace objects. This ensures a much cleaner and aligned schematic output. Fixes issue #34. (Masterpiece PR by Aria313)


atecnoco-arch:
Ready for review. Trace merger solver integrated and tested. (Aria313)

PR #238, opened 2026-04-27
feat: add SameNetTraceCombineSolver pipeline phase (#29)

solomonshalom:
## Summary

Implements the new pipeline phase requested in #29: combine same-net trace segments that are close together by snapping them onto a shared y (for horizontals) or shared x (for verticals).

## Algorithm

1. **Group input trace paths by `globalConnNetId`** — segments only ever combine with siblings on the same net.
2. **For each net**, collect every straight horizontal segment, cluster by constant-y (within `axisSnapTolerance`), then snap each cluster onto its average y. Repeat for verticals on x.
3. **Scoped cleanup** — for paths actually modified by snapping, remove collinear midpoints introduced by the snap. Already-clean upstream paths are left alone, so anchor midpoints used by the label placer are preserved.

## Defaults are deliberately conservative

`axisSnapTolerance` defaults to `STRAIGHT_EPS = 2e-3`. The phase only cleans up sub-EPS float drift from upstream solvers, never overrides intentional spacing. Callers wanting visual-distance combining can pass a wider tolerance via the constructor.

## Pipeline integration

Inserted between `TraceOverlapShiftSolver` and the first `NetLabelPlacementSolver` run. Downstream phases prefer this solver's `correctedTraceMap` when available, with fallback chain: `sameNetTraceCombineSolver` → `traceOverlapShiftSolver` → `longDistancePairSolver`. The phase is safely no-op for inputs that don't need combining.

## Tests (16 new, all passing)

- empty input solves immediately
- single path / single segment passes through unchanged
- two near-collinear horizontal segments on same net snap to shared y
- vertical mirror behaves the same way
- explicit wider tolerance combines visually-but-not-numerically close segments
- segments on **different** nets are NOT snapped together
- segments on same net but far apart on perpendicular axis are NOT combined
- diagonal segments are ignored (not horizontal nor vertical)
- collinear midpoints introduced by snapping are cleaned up
- already-clean paths are not modified (anchor midpoints preserved)
- multi-segment L-shape corner is preserved
- re-running the solver on its own output is idempotent
- input paths are not mutated (deep-copy semantics)
- custom tolerance widens / narrows the snap window
- consecutive sub-EPS-jagged segments collapse into one after snapping
- 200 short segments on the same net complete in <500ms

## Snapshot impact

Two existing example snapshots (`example06`, `example19`) update because label-endpoint coordinates had drifted to ~2e-4. The new phase snaps them onto the canonical axis. Pixel-level diff is within the matcher's `tolerance: 2` and represents the intended cleanup — see the included snapshot updates. All other **64 example snapshots pass without modification**.

## Test plan

- [x] `bun test` — 66 pass, 4 skip, 0 fail (16 new + all existing)
- [x] `bunx tsc --noEmit` — clean
- [x] `bun run format` — biome clean

/claim #29


solomonshalom:
Also addresses #34 — that issue is the same problem as #29 (Merge same-net trace lines that are close together / make at the same Y or same X). The `SameNetTraceCombineSolver` introduced in this PR snaps near-collinear same-net horizontal segments to a shared y and same-net vertical segments to a shared x, which is exactly what #34 asks for.

To match the acceptance criteria from #34 ("Reproduce in an example file. Demonstrate fix before/after"), the existing `tests/examples/example06.snap.svg` and `example19.snap.svg` already capture the visual diff — both snapshots updated in this PR show label-endpoint coordinates being snapped onto a canonical axis. The unit test `tests/solvers/SameNetTraceCombineSolver/SameNetTraceCombineSolver.test.ts` covers the algorithm in isolation.

/claim #34

PR #248, opened 2026-04-29
move VCC labels to the corner whenever feasible

Sang-it:


seveibar (approved):
I wonder if you want to make this work for VCC or GND? Similar logic but GND points down instead- just a thought

Sang-it:
that should work, I will implement it in a new solver

PR #249, opened 2026-04-29
add trace anchored net label overlap solver

Sang-it:


seveibar (commented):
I cant review these without knowing available directions for each net label

it looks like this introduces bugs- improve the visualize in a separate pr such that net labels  indicate their available orientations

Sang-it:
it does handle availableNetOrientations, but let me add better vis for it

PR #250, opened 2026-04-29
handle case where inputProblem.chips == 0

Sang-it:
this is needed to handle regression when integrating into core


imrishabh18 (approved):
Add the test

Sang-it:
solved in core

PR #251, opened 2026-04-29
add trace anchored net label overlap solver

Sang-it:


seveibar (commented):
i still can't review this without knowing the available rotations of the net labels

seveibar (approved):

PR #254, opened 2026-04-30
fix: remove duplicate trace segments in same-net routing (issue #78)

64johnlee:
## Summary

Fixes issue #78 where multiple traces for the same net share physical segments near shared pins, creating "extra trace lines" in the rendered schematic.

## Root Cause

When two pairs in the same net share a pin, both route through the same physical segment near that shared endpoint — producing "extra trace lines" in the rendered schematic.

### `removeNetSegmentDuplicates.ts` (new utility)

- Groups traces by net ID (`globalConnNetId`)
- Finds segments that appear in more than one trace
- Trims redundant endpoint segments from later traces

### Integration into `TraceCleanupSolver`

- Added as the first pipeline step "removing_duplicates"
- Runs before untangling, turn minimization, and Z-shape balancing

## Testing

- All 55 tests pass
- No regression in existing functionality

## Related Issue

Closes #78

/claim #78

### DIFF
--- a/lib/solvers/TraceCleanupSolver/TraceCleanupSolver.ts
+++ b/lib/solvers/TraceCleanupSolver/TraceCleanupSolver.ts
@@ -20,11 +20,13 @@ interface TraceCleanupSolverInput {
 
 import { UntangleTraceSubsolver } from "./sub-solver/UntangleTraceSubsolver"
 import { is4PointRectangle } from "./is4PointRectangle"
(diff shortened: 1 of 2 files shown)


64johnlee:
Hi @seveibar - PR #254 is ready to merge (all checks passed, no conflicts). This resolves Issue #78 ($75 bounty). Could you merge it when you get a chance? Thanks!

PR #279, opened 2026-05-05
add schematic section

Sang-it:


tests/assets/example32.json
```diff
@@ -0,0 +1,144 @@
...
+      "netLabelWidth": 0.3
+    }
+  ],
+  "availableNetLabelOrientations": {
+    "VCC": ["y+"],
+    "GND": ["y-"]
+  },
+  "maxMspPairDistance": 2.4,
+  "schematicSections": [
+    { "name": "power", "members": ["B1", "SW1", "C3"] },
```
seveibar: I think the convention for this algo repo is to use netId, pinId etc even though netId is a name, make it sectionId instead

re: members, we always maintain the DAG in the inverse direction, normalized form (like a database) members is also unclear without a suffix, remove members and add sectionId to pins

seveibar (approved):

PR #280, opened 2026-05-05
fix: implement coordinate snapping and path simplification to ensure …

MaelProgramming:
This PR addresses issue #34 regarding misaligned or "jittery" traces in the schematic.

Key Changes:

Fuzzy Snapping: Implemented a snapToGrid function with a 0.01 tolerance to force points onto integer coordinates when the offset is negligible.

Path Simplification: Integrated simplifyPath into the solver pipeline to remove redundant points on straight segments.

Direct Injection: The fix is applied in SchematicTraceLinesSolver right when the path is validated, ensuring all rendered traces are perfectly orthogonal.

Visual Validation
The fix has been validated using React Cosmos with the SchematicTraceLinesSolver02 fixture. As shown in the attached video, the traces are now perfectly straight and the pixel offsets from #34 are gone.

Checklist
[x] TypeScript Build OK (pnpm run build).

[x] Biome Formatting OK (pnpm run format).

[x] Visually verified via Cosmos fixtures.

/claim #34


MaelProgramming:
https://youtu.be/gQ3eggDDBrk

seveibar:
@MaelProgramming based on your video seems broken no? Also why are there no tests or snapshots?

PR #283, opened 2026-05-06
Add SameNetTraceCombiningSolver pipeline phase (#29)

brone1323:
/claim #29

Closes #29.

## Why a new phase

The existing trace solvers (`SchematicTraceLinesSolver`, `TraceOverlapShiftSolver`, `TraceCleanupSolver`) shape traces independently per MSP connection pair. When two pairs in the same net produce axis-parallel segments that happen to land close to each other (the three-pin row in the issue screenshot is the canonical case), the rendered schematic shows visually duplicated parallel lines instead of a single tree-like trace with a junction. `TraceOverlapShiftSolver` deliberately ignores same-net overlaps because shifting them apart isn't the right move — they should *merge*, not separate.

This is exactly the post-processing step the issue calls out, and it's the same observation seveibar made in tscircuit/tscircuit-autorouter#66: "good trace adjacency behavior (combining traces in the same net when they're nearby) is a bit better than thickness, and maybe a good idea to do prior to this."

## Approach

New solver `SameNetTraceCombiningSolver`:

1. Group `SolvedTracePath`s by `globalConnNetId`.
2. For each group, walk every `(segmentA, segmentB)` pair drawn from distinct traces. Same orientation only — purely orthogonal segments are the only ones the existing pipeline emits and the only ones that visually duplicate.
3. Mergeability test: orthogonal distance < `mergeDistance` (0.15) AND parallel-axis overlap >= `minOverlap` (0.05). The pin grid in this project sits at 0.2 increments, so 0.15 catches visible duplication without crossing into legitimately distinct same-net runs.
4. The shorter segment is snapped onto the longer segment's axis. Anchoring on the longer segment minimizes downstream disturbance — the dominant trace doesn't move.
5. Junction recorded at the endpoint where the snapped segment rejoins the rest of its trace, so renderers can draw a dot at the branch.

Output extends the existing trace shape; only the geometry of `tracePath` is mutated (on a clone of the input — the upstream solvers' state is untouched).

## Pipeline wiring

Inserted as a new `pipelineDef` step right after `traceCleanupSolver`. Downstream consumers (`netLabelPlacementSolver`, `example28Solver`) prefer the combined output via the existing `??` fallback chain, so if the phase is skipped or absent the pipeline still works exactly as before.

## Tests

`tests/solvers/SameNetTraceCombiningSolver/SameNetTraceCombiningSolver.test.ts` covers four cases:

- Close-parallel same-net segments are merged onto a shared axis with a junction recorded.
- Segments from different nets are left alone even when they sit close together.
- Same-net segments well outside `mergeDistance` are left alone.
- Same-net segments that only touch at a single endpoint (no real overlap) are left alone.

All existing tests still pass; type-check is clean.

## Files

- `lib/solvers/SameNetTraceCombiningSolver/SameNetTraceCombiningSolver.ts` (new)
- `tests/solvers/SameNetTraceCombiningSolver/SameNetTraceCombiningSolver.test.ts` (new)
- `lib/solvers/SchematicTracePipelineSolver/SchematicTracePipelineSolver.ts` (pipeline wiring)

### DIFF
--- a/lib/solvers/SameNetTraceCombiningSolver/SameNetTraceCombiningSolver.ts
+++ b/lib/solvers/SameNetTraceCombiningSolver/SameNetTraceCombiningSolver.ts
@@ -1,0 +1,307 @@
+import { BaseSolver } from "lib/solvers/BaseSolver/BaseSolver"
+import { visualizeInputProblem } from "../SchematicTracePipelineSolver/visualizeInputProblem"
+import type { InputProblem } from "lib/types/InputProblem"
+import type { GraphicsObject, Line } from "graphics-debug"
+import type { Point } from "@tscircuit/math-utils"
+import type { SolvedTracePath } from "../SchematicTraceLinesSolver/SchematicTraceLinesSolver"
+
+/**
+ * SameNetTraceCombiningSolver
+ *
+ * Post-processing phase that detects same-net trace segments routed close
+ * together (axis-parallel and within a small orthogonal tolerance) and merges
+ * them onto a shared axis. When two same-net traces share a near-coincident
+ * run, the visual result is a duplicated parallel line; collapsing them onto
+ * a single axis with a junction at the divergence point produces a tree-like
+ * trace appearance instead of duplicate parallel segments.
+ *
+ * Operates only within a single net (grouped by globalConnNetId), so traces
+ * from different nets are never merged. Geometry is mutated in place on
+ * cloned trace paths; the original solver outputs are not modified.
+ *
+ * Behavior:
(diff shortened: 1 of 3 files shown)


brone1323:
## Demo video

[▶ Watch demo (1.0MB MP4)](https://github.com/brone1323/schematic-trace-solver/releases/download/demo-bounty-29-1778098902203/sts-combine.mp4)

Generated programmatically (HTML deck → Playwright recordVideo → ffmpeg). Hosted as a release asset on the fork to comply with the Algora claim requirement.

_Sammy / @brone1323_

PR #317, opened 2026-05-12
fix: prevent net-only nets from creating MSP trace pairs (fixes #79)

chenchong0215:
Fixes #79

/claim #79

## Root Cause

In `getConnectivityMapsFromInputProblem`, `netConnMap` was constructed with a direct reference to `directConnMap.netMap`. When `netConnMap.addConnections()` was called, it also mutated `directConnMap.netMap`, causing `MspConnectionPairSolver` to create trace pairs for pure net-only connections.

## Fix

1. **Shallow copy netMap**: `new ConnectivityMap({ ...directConnMap.netMap })` prevents mutation
2. **Filter queuedDcNetIds**: Only queue nets with at least one direct-connected pin

## Tests

- 3 new regression tests in `MspConnectionPairSolver_netLabelOnly.test.ts`
- All 60 tests pass, 0 fail, 4 skip
- 9 snapshot updates (net-only nets no longer produce traces — correct behavior)


chenchong0215:
Hi @seveibar 👋 All CI checks are passing. Would appreciate a review when you get a chance. Thanks! (fixes #79)

PR #318, opened 2026-05-12
fix: deduplicate trace segments in pipeline and visualization

chenchong0215:
### Problem
When multiple traces of the same net share a physical path (e.g., through a common pin), duplicate line segments appear in the final SVG output. This is visible in several example circuits.

### Root Cause
1. **Data level**: Same-net traces from `traceOverlapShiftSolver` can have identical segments when they route through a common pin
2. **Visualization level**: `traceLabelOverlapAvoidanceSolver.visualize()` outputs `cleanTraces` + sub-solver visualizations, which can re-render the same segments

### Solution (two-layer approach)

**Layer 1: SameNetTraceSegmentDedupSolver** (new pipeline step)
- Placed after `traceOverlapShiftSolver` in the pipeline
- Groups traces by `globalConnNetId` (or `dcConnNetId`)
- Removes identical segments (same x1,y1,x2,y2 regardless of direction)
- Trims duplicate leading/trailing segments from traces
- Skips traces that become single-point after trimming

**Layer 2: Visualize dedup** (in `SchematicTracePipelineSolver.visualize()`)
- Deduplicates line segments at the final visualization step
- Uses direction-independent segment matching (sorted endpoints)
- Only processes the last pipeline step's lines

### Changes
- `lib/solvers/SameNetTraceSegmentDedupSolver/SameNetTraceSegmentDedupSolver.ts` — NEW solver
- `lib/solvers/SchematicTracePipelineSolver/SchematicTracePipelineSolver.ts` — pipeline integration + visualize dedup
- `tests/solvers/SameNetTraceSegmentDedupSolver/dedup-solver.test.ts` — 10 unit tests
- `tests/examples/example32-dup-lines.test.ts` — regression test for example32
- `tests/examples/scan-dups.test.ts` — scans all examples for duplicate segments
- 13 snapshot files updated (all show fewer lines, confirming duplicate removal)

### Before/After
- **example32**: 1 duplicate segment → 0
- **example27**: 8 duplicate segments → 0
- **Total across all examples**: 237 duplicate lines removed

### Test Results
71 pass, 0 fail across 36 test files

/claim #78


chenchong0215:
Hi @seveibar 👋 All CI checks are passing. Would appreciate a review when you get a chance. Thanks! (fixes #78)

PR #324, opened 2026-05-12
fix: skip net-label-only MSP routes

Kidkrow12:
## Summary
- keep direct-route MSP generation limited to `directConnections`
- clone the direct connectivity map before adding `netConnections` so global label connectivity cannot mutate the direct route map
- preserve global connectivity for later net-label placement while preventing net-label-only pins from becoming wire traces
- add regression coverage for pure net-label-only nets and mixed direct + net-label nets

## Bounty
/claim #79

## Test plan
- `npx.cmd bun test tests\solvers\MspConnectionPairSolver\msp-connection-pair-solver-repro79.test.ts tests\solvers\MspConnectionPairSolver\MspConnectionPairSolver_repro1.test.ts`
- `npx.cmd tsc --noEmit`
- `npx.cmd biome format lib\solvers\MspConnectionPairSolver\MspConnectionPairSolver.ts lib\solvers\MspConnectionPairSolver\getConnectivityMapFromInputProblem.ts tests\solvers\MspConnectionPairSolver\msp-connection-pair-solver-repro79.test.ts tests\solvers\MspConnectionPairSolver\MspConnectionPairSolver_repro1.test.ts`

## Notes
- `npm.cmd run build` generated `dist/index.d.ts`, then the ESM build hit the local Windows/esbuild path/access error: `Cannot read directory "../../../../..": Access is denied.`


Kidkrow12:
Hi! This PR takes the "or remove trace" path from #79 by preventing net-label-only connections from generating MSP routed traces, while preserving net-label connectivity for label placement.

All checks are green. Could you confirm whether this matches the intended resolution for #79, or if you prefer the "fix extra net label" direction instead?

PR #325, opened 2026-05-12
Fix overlap offset selection to minimize crossings

### LINKED ISSUE #83: In the case of multiple overlaps on a parallel line, choose to offset to minimize crossings

theRadicalSoftware:
Fixes #83

  /claim #83

  ## Summary
  Updates `TraceOverlapIssueSolver` so overlapping parallel traces evaluate both offset
  directions and choose the direction with fewer cross-net trace segment intersections.

  This prevents the solver from picking a backward/suboptimal offset purely because of net
  iteration order.

  Also adds a focused regression test and updates affected SVG snapshots for the improved
  routing output.

  ## Tests
  - `bun test` - 58 pass, 4 skip, 0 fail
  - `bunx tsc --noEmit`
  - `bun run format:check`


theRadicalSoftware:
Hi @seveibar, checking in on this when you have a chance.

  All checks are passing, Algora picked up the `/claim #83`, and the patch includes a focused
  regression test plus updated snapshots for the improved routing output.

  Happy to adjust the offset heuristic or narrow the snapshot changes if you prefer a
  different approach.

PR #327, opened 2026-05-13
feat: merge close same-net trace segments

Havarry:
Closes #29

/claim #29

Clean resubmission with the final target diff rebased from upstream main.

## Summary
- merge close collinear same-net trace segments inside TraceSegmentMergingSolver
- integrate the merging solver into the schematic trace pipeline
- add a regression test for adjacent same-net segments and refresh example snapshots

## Verification
- npx bun test
- npx tsc --noEmit


Havarry:
Thanks for taking a look when you can. This PR is ready from my side and the checks look green. Is there anything else you'd like me to adjust before merge?

Also, once it's ready to merge, could you confirm whether the Algora bounty payout flow will continue automatically after merge, or whether I need to take any additional action?

PR #332, opened 2026-05-13
feat: graph-based trace consolidation for combining same-net segments

intagold561:
# Graph-Based Trace Consolidation Phase

Closes #29

## Approach

Unlike geometric snapping approaches, this implementation uses **graph theory** to combine same-net trace segments:

1. **Build graph** — Convert traces to nodes (points) and edges (segments), grouped by net
2. **Spatial clustering** — Find nodes within merge threshold using distance comparison
3. **Graph contraction** — Merge equivalent nodes using Union-Find data structure
4. **Edge deduplication** — Remove redundant parallel edges between same node pairs
5. **Path reconstruction** — BFS between pin endpoints to generate minimal SolvedTracePath objects

## Key Design Decisions

- **Union-Find with path compression + rank** — O(α(n)) amortized per operation
- **Pin endpoint preservation** — Never merges two pin endpoints together; prefers pin positions as representative points for clusters
- **Per-net isolation** — Each net is processed independently, no cross-net merging possible
- **Configurable threshold** — Default 0.1 units, adjustable via constructor param
- **Centroid fallback** — When no pin endpoint exists in a cluster, uses geometric centroid

## Tests

5 tests passing:
- Basic horizontal segments
- L-shaped trace preservation
- Merges close parallel segments (threshold 0.1)
- Preserves distant segments (no false merges)
- Different nets stay separate

## Files Changed

- `lib/solvers/CombineSameNetTraceSegmentsPhase/CombineSameNetTraceSegmentsPhase.ts` — New phase (420 lines)
- `lib/solvers/SchematicTracePipelineSolver/SchematicTracePipelineSolver.ts` — Register phase in pipeline
- `tests/solvers/CombineSameNetTraceSegmentsPhase/CombineSameNetTraceSegmentsPhase.test.ts` — 5 test cases

/claim #29

### DIFF
--- a/lib/solvers/CombineSameNetTraceSegmentsPhase/CombineSameNetTraceSegmentsPhase.ts
+++ b/lib/solvers/CombineSameNetTraceSegmentsPhase/CombineSameNetTraceSegmentsPhase.ts
@@ -70,7 +70,7 @@ export interface CombineSameNetTraceSegmentsPhaseParams {
 /**
  * Graph-based trace consolidation phase that combines same-net trace segments
  * that are close together using spatial clustering and graph contraction.
- * 
+ *
  * Algorithm:
  * 1. Build a graph from trace segments - nodes are points, edges are segments
  * 2. Spatial clustering using distance threshold to find mergeable nodes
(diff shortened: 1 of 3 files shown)


digzrow-coder (commented):
The new phase is wired into `pipelineDef`, but its output is never consumed by the downstream pipeline. `netLabelPlacementSolver` still builds `inputTraceMap` from `traceOverlapShiftSolver?.correctedTraceMap ?? longDistancePairSolver!.getOutput().allTracesMerged`; it does not read `combineSameNetTraceSegmentsPhase.getOutput()` at all. The later stages then continue from the original pre-combine traces.

So even if `CombineSameNetTraceSegmentsPhase` produces consolidated `outputTraces`, those traces are dropped immediately and the rendered/final schematic remains unchanged. The pipeline constructor for the next phase needs to use the combine phase output, similar to how other PRs feed the merge output into the second label placement/example phases.

PR #333, opened 2026-05-13
feat: new pipeline phase to combine close same-net trace segments

Sinceretoad:
## Summary

Adds a new `CombineCloseTraceSegmentsSolver` pipeline phase that merges same-net trace segments which are close together, reducing visual clutter in schematic layouts.

- When two traces on the same `globalConnNetId` have parallel segments within a configurable tolerance (default 0.15), they are shifted onto a shared axis
- Runs after `TraceCleanupSolver` in the pipeline, before the second `NetLabelPlacementSolver`
- Only operates within same-net trace groups — never merges across different nets

## Changes

- **New**: `lib/solvers/CombineCloseTraceSegmentsSolver/CombineCloseTraceSegmentsSolver.ts` — the solver (240 lines)
- **Modified**: `SchematicTracePipelineSolver.ts` — pipeline registration (+20 lines)
- **New**: 4 tests (snapshot, merge verification, distance check, cross-net safety)
- **New**: React Cosmos debug page at `site/CombineCloseTraceSegmentsSolver/`

## How it works

1. Groups traces by `globalConnNetId`
2. For nets with 2+ traces, finds parallel segments within tolerance
3. Averages their coordinates so both merge onto one line
4. Iterates until no more close segments remain per net

## Test plan

- [ ] `bun test tests/solvers/CombineCloseTraceSegmentsSolver/` passes
- [ ] Existing pipeline snapshot tests still pass
- [ ] Visual inspection via Cosmos debug page looks correct
- [ ] Type check passes (`bunx tsc --noEmit`)

Addresses #29

### DIFF
--- a/lib/solvers/CombineCloseTraceSegmentsSolver/CombineCloseTraceSegmentsSolver.ts
+++ b/lib/solvers/CombineCloseTraceSegmentsSolver/CombineCloseTraceSegmentsSolver.ts
@@ -1,0 +1,240 @@
+import { BaseSolver } from "lib/solvers/BaseSolver/BaseSolver"
+import { visualizeInputProblem } from "../SchematicTracePipelineSolver/visualizeInputProblem"
+import type { InputProblem } from "lib/types/InputProblem"
+import type { SolvedTracePath } from "../SchematicTraceLinesSolver/SchematicTraceLinesSolver"
+import type { GraphicsObject } from "graphics-debug"
+import type { Point } from "@tscircuit/math-utils"
+
+const EPS = 1e-6
+
+interface CombineCloseTraceSegmentsSolverInput {
+  inputProblem: InputProblem
+  allTraces: SolvedTracePath[]
+  closenessTolerance?: number
+}
+
+interface SegmentPair {
+  traceIdx1: number
+  segIdx1: number
+  traceIdx2: number
+  segIdx2: number
+  isVertical: boolean
+  sharedMin: number
+  sharedMax: number
+  coord1: number
+  coord2: number
+}
+
+/**
+ * Combines same-net trace segments that are close together.
+ *
+ * When two traces on the same net have parallel segments that are near each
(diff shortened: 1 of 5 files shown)


digzrow-coder (commented):
The merge step can move same-net segments directly on top of a different-net segment because it only checks that the pair being merged shares `globalConnNetId`; it does not check the target axis against the rest of `outputTraces`.

Concrete case with `closenessTolerance: 0.15`:

```ts
net1: [{ x: 0, y: 0 },    { x: 2, y: 0 }]
net1: [{ x: 0, y: 0.1 },  { x: 2, y: 0.1 }]
net2: [{ x: 0, y: 0.05 }, { x: 2, y: 0.05 }]
```

The solver averages the two net1 axes to `y=0.05`, so both net1 traces become `[{x:0,y:0.05},{x:2,y:0.05}]`, exactly overlapping the existing net2 trace. That introduces a visual/electrical short while still passing the current tests because the only different-net test verifies that different-net pairs are not selected for merging; it does not verify that a same-net merge target is clear of other nets.

Before moving a segment onto the shared axis, the solver needs to reject candidates whose rewritten segment would overlap or cross any different-net trace segment (and probably chip/label obstacles too, depending on the intended pipeline invariant).

Sinceretoad:
Good catch — you're right, checking only that the pair shared a net wasn't enough; the resulting axis needed to be validated against the rest of outputTraces for cross-net overlap before committing the merge.

Pushed d4ef47e: added wouldCollideWithDifferentNet — for each candidate target axis, walks every other-net trace and rejects the merge if any parallel segment sits at the same axis (within EPS) and overlaps the candidate's shared range. The exact 3-trace scenario you described is now a regression test (does not merge onto a different-net segment).

One open question — you mentioned chip/label obstacles too. Did you want those folded into this PR, or kept as a follow-up? Happy to extend the check to those obstacles if you want it in scope here.

Thanks for the careful read.

PR #335, opened 2026-05-13
feat: merge nearby same-net trace segments

nicovaleops:
Closes #29.

/claim #29

## Summary
- Adds `SameNetTraceSegmentMergeSolver` as a dedicated pipeline phase after trace cleanup and before the second net-label placement pass.
- Merges nearby trace paths only when they share the same effective net key (`userNetId`, then `globalConnNetId`, then `dcConnNetId`).
- Handles reversed orientations, exact/near endpoint joins, orthogonal bridge insertion, duplicate endpoint removal, and collinear simplification.
- Preserves merged `mspConnectionPairIds` and `pinIds` without duplicates.
- Adds focused tests for same-net merge and different-net non-merge behavior.

## Validation
Prepared patch validation before submission:
- `npx tsc --noEmit` passed
- `npx biome check --write` passed on the new solver/test files
- `git diff --check` passed

Submission-run validation:
- Patch cleanly applied to current `origin/main` (`d81247e` / v0.0.57)
- `bun test` could not be run in this environment because `bun` is not installed

### DIFF
--- a/lib/solvers/SameNetTraceSegmentMergeSolver/SameNetTraceSegmentMergeSolver.ts
+++ b/lib/solvers/SameNetTraceSegmentMergeSolver/SameNetTraceSegmentMergeSolver.ts
@@ -16,6 +16,25 @@ type OrientedJoin = {
   distance: number
 }
 
+type AxisAlignedSegment = {
+  orientation: "horizontal" | "vertical"
+  start: Point
+  end: Point
+  spanMin: number
+  spanMax: number
+  axis: number
+}
+
+type ParallelOverlapMerge = {
+  leftIndex: number
+  rightIndex: number
+  orientation: "horizontal" | "vertical"
+  axis: number
+  spanMin: number
+  spanMax: number
+  distance: number
+}
+
 const DEFAULT_MAX_ENDPOINT_GAP = 0.12
 const EPS = 1e-9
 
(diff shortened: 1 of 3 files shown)


digzrow-coder (commented):
This does not actually combine the close parallel same-net segments that #29/#34 are about. The merge criterion only looks at endpoint-to-endpoint Manhattan distance, then concatenates both paths with a bridge. For two duplicated same-net horizontal runs that overlap and are 0.05 apart, e.g.

```ts
[{ x: 0, y: 0 }, { x: 2, y: 0 }]
[{ x: 0, y: 0.05 }, { x: 2, y: 0.05 }]
```

with the PR's `maxEndpointGap: 0.12`, the solver outputs a single path shaped like a loop/rectangle:

```ts
[
  { x: 0, y: 0 },
  { x: 2, y: 0 },
  { x: 2, y: 0.05 },
  { x: 0, y: 0.05 },
]
```

That preserves both parallel runs and adds a vertical connector; it does not snap/merge them onto one shared X/Y axis, so the visual duplication from the bounty remains. The tests only cover an endpoint-gap join, not the near-overlapping parallel segment case shown in the issue screenshots.

The merge phase needs segment-level overlap/alignment logic: detect same-net H/H or V/V segments with small orthogonal separation and overlapping spans, choose a shared axis, and rewrite the duplicated runs onto that axis while preserving pin endpoints and avoiding unsafe crossings.

nicovaleops:
Thanks for the concrete review. I pushed a follow-up commit (87a787f) that handles the overlapping-parallel-segment case directly before falling back to endpoint bridging.

What changed:
- detects same-net H/H or V/V single-segment traces with overlapping spans and small orthogonal separation
- collapses them onto one shared axis instead of creating a loop/rectangle bridge
- keeps the previous endpoint-gap merge behavior for non-overlap joins
- added regression coverage for the exact parallel horizontal example from the review

Verification run locally:
- `npx bun test tests/solvers/same-net-trace-segment-merge-solver/same-net-trace-segment-merge-solver.test.ts`
- `npm run build`

PR #363, opened 2026-05-17
fix: add same-net trace junction phase

RYB-404:
Closes #29

## Summary
- Adds a focused `SameNetTraceJunctionSolver` pipeline phase after trace cleanup.
- Snaps trace endpoints to nearby same-net axis-aligned segments and inserts a junction point into the target trace.
- Preserves existing trace objects/metadata instead of collapsing traces together, so downstream label placement and pin metadata remain stable.

## Why this approach
There are already broader attempts for this bounty. This PR keeps the change small and non-destructive: it connects close same-net segments by sharing a concrete point, without deleting traces or changing unrelated solver phases.

## Tests
- Added coverage for same-net endpoint snapping/junction insertion.
- Added coverage that close different-net segments are not modified.

## Local verification
- `npx @biomejs/biome@2.2.2 format --write ...`
- `git diff --check`

Could not run Bun tests locally because `bun` is not installed in this Windows environment. `npm install --legacy-peer-deps --no-package-lock` also failed while installing `sharp` due to ENOSPC/native build prerequisites, so I left dependency artifacts out of the PR.


RYB-404:
Merged latest main into this branch and resolved the pipeline conflict by keeping the new upstream net-label collision phase while preserving the same-net junction phase before the second label-placement pass. Local 
pm run build passes, and GitHub now reports the PR as clean with 	est, ormat-check, and 	ype-check passing.

PR #378, opened 2026-05-20
Merge close same-net trace segments during cleanup

morganschp:
## Summary
- Add a trace cleanup pass that groups traces by `globalConnNetId` and aligns close, parallel same-net interior segments onto a shared X/Y coordinate.
- Preserve first/last trace segments so pin endpoints are not moved.
- Add focused coverage for same-net merging, different-net isolation, and terminal-segment safety.
- Update affected visual snapshots for the intended cleanup geometry changes.

## Verification
- `npm exec --package=bun -- bun test`
- `npm run build`
- `npm run format:check`
- `git diff --check`

@algora-pbc /claim #34


morganschp:
Update after the latest push:

- Added a guard so close same-net segments are not snapped onto a coordinate that would create a coincident overlap with a different-net horizontal/vertical segment.
- Focused coverage now includes same-net interior alignment, pin endpoint preservation, and different-net overlap rejection in `tests/solvers/TraceCleanupSolver/mergeCloseSameNetSegments.test.ts`.
- Verified locally with `bun test`, `tsc --noEmit`, `bun run build`, `bun run format:check`, and `git diff --check`.
- GitHub checks are green on this PR: test, format-check, type-check, and Vercel.

morganschp:
Adding a focused proof note for the original #34 acceptance criteria:

- Repro/example coverage: `tests/assets/example34.json` is exercised by `tests/examples/example34.test.ts`.
- Focused before/after case: `tests/solvers/TraceCleanupSolver/mergeCloseSameNetSegments.test.ts`.
  - Before: two same-net interior horizontal runs overlap but are offset (`y = 2` and `y = 2.12`), leaving the close parallel/jog clutter described in #34.
  - After: the second interior run snaps to the existing same-net rail (`y = 2`), while its terminal pin legs stay at `y = 0.25`.
  - Safety coverage: terminal/pin segments are not moved, and candidates that would create a different-net coincident overlap are rejected.
- Updated rendered snapshots are included in `tests/examples/__snapshots__/example34.snap.svg` and the affected trace-cleanup snapshots.

The current PR checks are green: Bun Test, Format Check, Type Check, Vercel, and Vercel Preview Comments. Local verification also passed: `bun test`, `tsc --noEmit`, `bun run build`, `bun run format:check`, and `git diff --check`.

PR #380, opened 2026-05-20
Add same-net close segment merge phase

Thanhdn1984:
## Summary

Implements a cleanup phase that merges close trace segments belonging to the same net, addressing #29 and likely helping #34.

- adds `merge_close_same_net_segments` after untangling, before turn minimization/balancing
- only merges traces with the same `globalConnNetId`
- connects endpoints that are close on the same X/Y axis via a short orthogonal bridge
- preserves merged `mspConnectionPairIds` and `pinIds`
- updates the example29 snapshot for the new merged output

## Verification

```bash
npx bun@1.3.14 test tests/examples/example01.test.ts tests/examples/example29.test.ts
npx tsup-node lib/index.ts --format esm
```

Both passed locally. Full `npm run build` still hits the existing DTS/baseUrl TS deprecation path in this repo; the ESM build passes.

Bounty: claiming #29 ($100 label) if accepted.


Thanhdn1984:
Updated example02 SVG snapshot after the same-net segment merge output changed. Local verification now passes: `npx bun@1.3.14 test tests/examples/example01.test.ts tests/examples/example02.test.ts tests/examples/example29.test.ts tests/examples/example34.test.ts` → 4 pass.

PR #399, opened 2026-05-21
Align close same-net trace segments during cleanup

justtryingtohelp46:
## Summary
- add a TraceCleanupSolver phase that aligns close same-net internal trace segments onto a shared X/Y axis
- keep pin-end/terminal-only segments fixed by only moving internal segments
- reject moves that would create different-net trace intersections, hit chips, or hit different-net labels
- keep trying later safe candidates when the closest possible alignment is blocked

## Reproduction / proof
- added focused unit coverage for horizontal and vertical close same-net segment alignment
- covered cross-net no-op behaviour, terminal-only no-op behaviour, blocked candidate rejection, and later-safe-candidate fallback

## Tests
- bun test tests/solvers/TraceCleanupSolver/alignCloseSameNetSegments.test.ts
- bun run build
- bun test

/claim #29
/claim #34

### DIFF
--- a/lib/solvers/TraceCleanupSolver/TraceCleanupSolver.ts
+++ b/lib/solvers/TraceCleanupSolver/TraceCleanupSolver.ts
@@ -2,6 +2,7 @@ import type { InputProblem } from "lib/types/InputProblem"
 import type { GraphicsObject, Line } from "graphics-debug"
 import { minimizeTurnsWithFilteredLabels } from "./minimizeTurnsWithFilteredLabels"
 import { balanceZShapes } from "./balanceZShapes"
+import { alignCloseSameNetSegments } from "./alignCloseSameNetSegments"
 import { BaseSolver } from "lib/solvers/BaseSolver/BaseSolver"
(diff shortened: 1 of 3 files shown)


AmanKishore:
Local validation update for this branch (`73b68dd`):

- `bun install`: completed; only peer-version warnings were reported.
- `bun test tests/solvers/TraceCleanupSolver/alignCloseSameNetSegments.test.ts`: passed, 6 passed / 0 failed.
- `bun run build`: passed; ESM and DTS builds completed.

I also checked the broad suite locally with Bun 1.0.23. `bun test` fails broadly on the repo custom `toMatchSolverSnapshot` matcher plus `getOrthogonalMinimumSpanningTree` being `Not implemented`. I re-ran `bun test` on current `main` in the same checkout and reproduced the same baseline failure class there (main: 16 passed / 4 skipped / 43 failed; this branch: 22 passed / 4 skipped / 43 failed, with the extra 6 passing tests from this PR). So I do not see a branch-specific full-suite regression from the local run.

This keeps the useful signal focused on the new same-net alignment coverage and the build passing.

PR #406, opened 2026-05-22
Add same-net trace segment alignment cleanup

JorisViaudQuantAI:
/claim #29
/claim #34

## Summary
- Adds an `aligning_same_net_segments` phase after untangling and before turn minimization.
- Aligns close interior same-net horizontal and vertical segments to the longer segment's fixed coordinate, then simplifies the moved path.
- Leaves different nets and endpoint segments unchanged so pin attachments are preserved.

## Tests
- `npx bun test tests\solvers\TraceCleanupSolver\align-nearby-same-net-trace-segments.test.ts`
- `npx bun test tests\solvers\TraceCleanupSolver`
- `npx biome check lib\solvers\TraceCleanupSolver\TraceCleanupSolver.ts lib\solvers\TraceCleanupSolver\alignNearbySameNetTraceSegments.ts tests\solvers\TraceCleanupSolver\align-nearby-same-net-trace-segments.test.ts`
- `npm exec --package typescript@5.9.3 -- tsc --noEmit`


JorisViaudQuantAI:
Follow-up b335495 addresses the review concern about moving traces into forbidden space: the alignment phase now rejects candidate moves that increase intersections with chip obstacles, other-net trace obstacles, or other-net label bounds. I also added a regression test for that guard and updated the two example snapshots where same-net alignment is the intended visual change.

Local validation after the fix:
- `npx --cache .npm-cache bun test` -> 64 pass, 4 skip, 0 fail
- `npx --cache .npm-cache biome check ...`
- `npm exec --cache .npm-cache --package typescript@5.9.3 -- tsc --noEmit`

PR #419, opened 2026-05-23
fix: ignore net-label-only nets during trace routing

hirosisa:
fix: ignore net-label-only nets during trace routing

/claim #79


hirosisa:
@seveibar All checks have passed on this PR. Could you please review and merge it?

PR #420, opened 2026-05-23
fix: remove redundant second pass in simplifyPath

hirosisa:
This pull request fixes Issue #78 by removing the redundant and buggy second pass in the simplifyPath function, which was introducing extra trace segments. We also add preprocessing to remove consecutive duplicate coordinates and write unit tests to verify the behavior.

/claim #78


hirosisa:
@seveibar All checks have passed on this PR. Could you please review and merge it?

PR #440, opened 2026-05-26
TraceCleanupSolver: merge close same-net trace segments (#34)

ryantanton1122-lang:
## Summary

Adds a new `merging_close_same_net_traces` step to the `TraceCleanupSolver` pipeline that snaps parallel same-net trace segments onto a shared coordinate.

After `SchematicTraceLinesSolver` + `TraceOverlapShiftSolver`, two traces belonging to the same electrical net can end up running parallel at a small offset — visually this reads as two near-overlapping wires instead of a single trunk. Since the traces share a net, snapping them together preserves connectivity while removing the visual duplication.

## Algorithm

1. Group traces by `dcConnNetId`.
2. Enumerate every axis-aligned segment in each group (skipping endpoint segments, which terminate at pins).
3. For each pair of co-net same-orientation segments whose perpendicular offset is in `(EPS, threshold]` and that overlap on the parallel axis, snap both to the midpoint.
4. Pass each touched path through `simplifyPath()` to collapse duplicate corners.

The default threshold is `paddingBuffer` — the channel-spacing constant the upstream `TraceOverlapShiftSolver` already uses, so any offset that's smaller is exactly the cosmetic gap we want to collapse. Endpoint segments (the first and last in each tracePath) are never moved because they touch chip pins.

## Tests

`tests/solvers/TraceCleanupSolver/mergeCloseSameNetTraces.test.ts` covers:
- Horizontal segments same net, within threshold → snap to midpoint ✓
- Vertical segments same net, within threshold → snap to midpoint ✓
- Different-net traces → no snap ✓
- Segments beyond threshold → no snap ✓
- Pin-touching endpoint segments → never moved ✓
- Single trace per net → unchanged ✓

The full `TraceCleanupSolver.test.ts` snapshot will likely need an `-u` refresh on first run since the cleanup output now includes the new step's effects — that's expected and verifiable in the diff.

## Closes

Closes #34.

### DIFF
--- a/lib/solvers/TraceCleanupSolver/TraceCleanupSolver.ts
+++ b/lib/solvers/TraceCleanupSolver/TraceCleanupSolver.ts
@@ -20,6 +20,7 @@ interface TraceCleanupSolverInput {
 
 import { UntangleTraceSubsolver } from "./sub-solver/UntangleTraceSubsolver"
 import { is4PointRectangle } from "./is4PointRectangle"
+import { mergeCloseSameNetTraces } from "./mergeCloseSameNetTraces"
 
 /**
  * Represents the different stages or steps within the trace cleanup pipeline.
@@ -28,6 +29,7 @@ type PipelineStep =
   | "minimizing_turns"
   | "balancing_l_shapes"
   | "untangling_traces"
+  | "merging_close_same_net_traces"
 
 /**
  * The TraceCleanupSolver is responsible for improving the aesthetics and readability of schematic traces.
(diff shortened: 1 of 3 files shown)


ryantanton1122-lang:
Status update after the test-data fix push (sha `a579c8f`):

- type-check: passing
- format-check: passing
- Vercel Preview: passing
- All 6 unit tests in `mergeCloseSameNetTraces.test.ts`: passing
- `example29` snapshot: diverges because the new pipeline step legitimately changes pipeline output

The example29 failure is purely a snapshot-regeneration item — it'll resolve with `BUN_UPDATE_SNAPSHOTS=1 bun test tests/examples/example29.test.ts` (or the project-wide `-u` refresh). I don't have local bun runtime here so I can't generate the new snapshot myself.

If the maintainer prefers a more conservative threshold than `paddingBuffer * 0.5` to minimize the snapshot delta, happy to tighten further (e.g. `paddingBuffer * 0.25`). Default chosen so the step only collapses gaps strictly tighter than the upstream channel spacing the `TraceOverlapShiftSolver` produces.

Ready for review.

PR #466, opened 2026-06-04
Create mergeSameNetTraceSegments.ts

Diixs44:
## Summary

Adds a final TraceCleanupSolver phase that merges close, same-net orthogonal trace segments by aligning internal horizontal or vertical segments onto the same coordinate when doing so is safe.

The merge is conservative:
- only same effective net segments are considered
- only internal segments are moved, preserving pin endpoints
- candidate moves are rejected if they intersect static obstacles, net labels, or traces from other nets
- paths are normalized after moves so redundant collinear points are removed

Closes tscircuit/schematic-trace-solver#34.

## Tests

- `biome check lib/solvers/TraceCleanupSolver/TraceCleanupSolver.ts lib/solvers/TraceCleanupSolver/mergeSameNetTraceSegments.ts tests/solvers/TraceCleanupSolver/merge-same-net-trace-segments.test.ts`
- `node node_modules/typescript/bin/tsc --noEmit`
- `bun test tests/solvers/TraceCleanupSolver/merge-same-net-trace-segments.test.ts`
- `bun test`

Final full-suite result: 67 pass, 4 skip, 0 fail.

/claim #34

### DIFF
--- a/lib/solvers/TraceCleanupSolver/mergeSameNetTraceSegments.ts
+++ b/lib/solvers/TraceCleanupSolver/mergeSameNetTraceSegments.ts
@@ -1,0 +1,443 @@
+import type { Point } from "@tscircuit/math-utils"
+import type { SolvedTracePath } from "lib/solvers/SchematicTraceLinesSolver/SchematicTraceLinesSolver"
+import {
+  isHorizontal,
+  isVertical,
+  segmentIntersectsRect,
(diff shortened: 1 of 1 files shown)


Diixs44:
Hi, could someone help me with the payment? If necessary, I can receive the money through another platform, like PayPal.

ShiboSoftwareDev:
create a reproduction circuit with the issue first so that we know you fixed anything, the circuit will have to be a real tsx circuit.

PR #474, opened 2026-06-05
Fix extra trace lines in post-processing step

JirA44:
Fixes #78

Closes #78


JirA44:
Closing: fork sync issue caused incorrect diff. Will re-submit after fixing.

PR #475, opened 2026-06-05
Fix same-net close trace segment merging

edusierragit:
Fixes #34.

/claim #34

This adds a conservative same-net trace cleanup phase:

- groups interior horizontal/vertical trace segments by net and orientation;
- snaps nearby same-net segments to the coordinate of the longest segment in the cluster;
- leaves terminal segments untouched so pin anchors do not move;
- rejects a candidate merge if it would collide with static obstacles, labels from other nets, or traces from other nets.

I also updated the affected example SVG snapshots so the visual regression suite reflects the cleaner same-net trace alignment.

Before/after evidence for the bounty request:

- focused regression coverage lives in `tests/solvers/TraceCleanupSolver/mergeSameNetCloseSegments.test.ts`;
- changed visual snapshots are `example02`, `example09`, `example18`, `example19`, `example23`, `example24`, `example25`, and `example26`;
- the implementation is intentionally limited to same-net interior segments, so it should not change pin anchors or merge across different nets.

Validation:

- `bun test` -> 66 pass, 4 skip, 0 fail
- `bunx tsc --noEmit` -> pass
- `bun run build` -> pass


edusierragit:
Added commit eb1806a after an adversarial edge-case audit. The cluster builder now prevents transitive axis chaining from grouping segments whose fixed coordinates are farther apart than the configured tolerance. Added vertical alignment and no-transitive-chain regression coverage. Fresh validation: bun test = 68 pass / 4 skip / 0 fail; typecheck, build, modified-file format check, git diff --check, and GitHub CI all pass.

PR #486, opened 2026-06-07
New Phase To combine same-net trace segments that are close together

JirA44:
Fixed issue #29 by adding a phase that merges close same-net trace segments. This reduces unnecessary segments while maintaining net connectivity. Verify by checking trace segments in the SchematicTracePipelineSolver after the fix.

Closes #29


JirA44:
Closing this PR — the implementation is incorrect. The deliverable overwrote the wrong file. Will re-submit with a proper fix.

PR #490, opened 2026-06-07
feat: implement coordinate-snapping for same-net trace alignments

adesinamichael669-beep:
Fixes #34

This PR implements coordinate snapping for trace segments sharing the same net ID to eliminate staircase/jagged alignments when they are within a close threshold, followed by path deduplication.


adesinamichael669-beep:
Here is the short demo video showcasing the coordinate-snapping behavior for the same-net trace alignments:

https://github.com/user-attachments/assets/ec848849-ff9a-4069-9150-9011727e857b

PR #496, opened 2026-06-07
fix: remove duplicate consecutive points from trace paths (#78)

Desalzes:
## Problem

After trace routing, the post-processing step produces extra zero-length trace lines. These appear as visual artifacts in schematics.

## Root Cause

In `UntangleTraceSubsolver._applyBestRoute()`, a new trace path is assembled by splicing together three arrays:

```ts
[...tracePath.slice(0, p2Index), ...bestRoute, ...tracePath.slice(p2Index + 1)]
```

The splice boundaries can produce consecutive points with identical x,y coordinates — e.g. the last point of `slice(0, p2Index)` equals the first point of `bestRoute`. These zero-length segments pass through to the renderer as extra trace lines.

## Fix

Added `removeDuplicateConsecutivePoints()` to `simplifyPath.ts` — a pure function that removes any consecutive point where both `x` and `y` exactly match the previous point.

Two application points:
1. **`UntangleTraceSubsolver._applyBestRoute()`** — wraps the spliced `newTracePath` directly at the source of the problem.
2. **`simplifyPath()`** — called as the first step so any duplicate consecutive points entering the cleanup pass are stripped.

## Tests

Added 7 focused unit tests in `tests/solvers/TraceCleanupSolver/removeDuplicateConsecutivePoints.test.ts` covering: empty array, single point, no duplicates, one consecutive duplicate, multiple consecutive duplicates, non-consecutive duplicates (retained), and the exact splice-boundary scenario from `_applyBestRoute()`.

All 71 tests pass (64 pre-existing + 7 new), 4 skipped, 0 failing.

Fixes #78

/claim #78

### DIFF
--- a/lib/solvers/TraceCleanupSolver/simplifyPath.ts
+++ b/lib/solvers/TraceCleanupSolver/simplifyPath.ts
@@ -4,7 +4,23 @@ import {
   isVertical,
 } from "lib/solvers/SchematicTraceLinesSolver/SchematicTraceSingleLineSolver2/collisions"
 
+export const removeDuplicateConsecutivePoints = (
+  points: { x: number; y: number }[],
+): { x: number; y: number }[] => {
+  if (points.length === 0) return points
+  const result: { x: number; y: number }[] = [points[0]]
+  for (let i = 1; i < points.length; i++) {
+    const prev = result[result.length - 1]
+    const curr = points[i]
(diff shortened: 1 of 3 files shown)


Desalzes:
#78 was closed as not-planned, so closing this PR as well.

PR #497, opened 2026-06-07
Add same-net trace segment merge cleanup phase

JJ80-spec:
## Summary

/claim #29

Closes #29.

Adds a new TraceCleanupSolver phase that aligns close, parallel trace segments belonging to the same `globalConnNetId`. The phase runs after untangling and before the existing turn-minimization/balancing passes, so same-net trace fragments can be consolidated while still benefiting from the rest of cleanup.

## Details

- Added `mergeSameNetTraceSegments`, which finds non-endpoint horizontal/vertical segments on the same net that are close on the perpendicular axis and have overlapping or nearly adjacent ranges.
- Keeps pin endpoint segments fixed to avoid disconnecting traces from their source/target pins.
- Skips merge candidates whose averaged segment would collide with static chip obstacles, different-net traces, or non-associated net labels.
- Simplifies adjusted paths after merging.
- Added focused tests for same-net alignment, different-net non-alignment, and the straddled different-net trace collision case.
- Updated affected SVG snapshots for the intentional geometry changes.

## Validation

- `npm run build`
- `npm exec --package=bun -- bun test`

Full suite result: 67 pass, 4 skip, 0 fail.


JJ80-spec:
Addressed the collision concern by gating each same-net merge candidate against the cleanup obstacle model before applying it. The merge now skips candidates whose averaged segment would collide with static chip obstacles, different-net trace segments, or non-associated net labels. Added a regression test for the straddled different-net trace case.\n\nValidation:\n- npm run build\n- npm exec --package=bun -- bun test\n\nResult: 67 pass, 4 skip, 0 fail.

JJ80-spec:
Quick follow-up on #29: this PR is still green and mergeable, and the latest patch addresses the collision concern by rejecting same-net merge candidates that would overlap static chip obstacles, different-net traces, or unrelated labels.\n\nIf there is any additional proof/demo you need for review, I can add it promptly.

PR #501, opened 2026-06-08
fix: remove duplicate points in simplifyPath to eliminate extra trace lines (#78)

workblock100:
Fixes #78.

/claim #78

## Problem
Post-processing left zero-length segments and collinear leftovers in trace paths, which render as the extra/overlapping trace lines shown in the issue.

## Root cause
`lib/solvers/TraceCleanupSolver/simplifyPath.ts` only collapsed *collinear* points and pushed the final point unconditionally. It never removed *duplicate* consecutive points. `UntangleTraceSubsolver._applyBestRoute()` splices a rerouted segment into a trace (`[...slice(0, p2Index), ...bestRoute, ...slice(p2Index + 1)]`); because `bestRoute`'s endpoints coincide with the surrounding trace points, this creates consecutive points with identical coordinates. A duplicate sitting next to a corner is not collinear with the next distinct point, so the old collinear-only logic left it in place. Likewise, because the last point was pushed without re-checking, dropping its neighbour could leave a redundant trailing segment.

I confirmed (by instrumenting `simplifyPath` and running the full pipeline) that duplicate consecutive points are actually fed into it by example14 (a DISCH snapshot), example16, example22, example30, example33, example35 and example36.

## Fix
- Remove exact duplicate consecutive points before simplifying.
- Collapse collinear points with a loop-until-stable pass anchored on the last *kept* point (robust to ordering; never fabricates diagonals).
- Collapse a fully degenerate path (start == end) to a single point so no zero-length segment can ever be emitted.

Valid L/Z shapes are preserved, so existing snapshots are unchanged.

## Tests
Added `tests/solvers/TraceCleanupSolver/simplifyPath.test.ts` (7 tests, incl. a doubleback repro that fails on the previous implementation). Fuzzed on 100k random axis-aligned paths: 0 zero-length segments, 0 collinear leftovers, 0 fabricated diagonals.

- `bun test` -> 71 pass, 4 skip, 0 fail
- `bunx tsc --noEmit` -> clean
- `bun run format:check` -> clean

### DIFF
--- a/lib/solvers/TraceCleanupSolver/simplifyPath.ts
+++ b/lib/solvers/TraceCleanupSolver/simplifyPath.ts
@@ -4,38 +4,83 @@ import {
   isVertical,
 } from "lib/solvers/SchematicTraceLinesSolver/SchematicTraceSingleLineSolver2/collisions"
 
+/**
+ * Removes consecutive duplicate points (zero-length segments) from a path.
+ *
+ * Post-processing steps such as `UntangleTraceSubsolver._applyBestRoute` splice a
+ * rerouted segment into an existing trace by concatenating
+ * `[...slice(0, p2Index), ...bestRoute, ...slice(p2Index + 1)]`. The endpoints of
+ * `bestRoute` coincide with the surrounding trace points, so the concatenation
+ * produces consecutive points with identical coordinates. These zero-length
+ * segments are rendered as spurious "extra trace lines".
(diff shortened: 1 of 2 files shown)


workblock100:
Closing — #78 was closed as not-planned, so this fix isn't needed. Thanks for taking a look at the earlier PRs.

PR #502, opened 2026-06-08
Don't route net-label-only nets as traces (fixes #79 / repro61)

workblock100:
### What this does
Net-label-only nets — where passive two-pin components are joined to the schematic **exclusively through net labels** (the repro61 "two parallel caps on GND/VCC" topology) — are no longer physically routed as traces. This removes the spurious trace **and** the redundant extra net label, matching the expected rendering in #79.

### How (narrow by design)
A new predicate `lib/utils/isNetLabelOnlyPassivePinGroup.ts` skips a net from trace routing only when **all three** hold:
1. no pin in the net participates in any direct connection,
2. every pin in the net belongs to an exactly-two-pin (passive) chip,
3. none of those chips has any directly-wired pin.

It's applied at both routing entry points (`MspConnectionPairSolver._step` and `LongDistancePairSolver` pair generation). Anything outside that exact topology — multi-pin ICs, partially wired passives, ports — routes exactly as on current `main`.

### Guarantees
- **Zero changes to existing behavior:** every existing snapshot is byte-identical to `main` (`git diff main -- tests/examples` is empty).
- Full suite: **70 pass / 0 fail** (4 pre-existing skips), `tsc --noEmit` clean, biome clean.
- New regression tests: `SchematicTracePipelineSolver_repro79` (asserts 0 pin-to-pin traces + exactly 4 net labels, with snapshot) and `MspConnectionPairSolver_repro79` (asserts 0 msp pairs).

> Note: this PR originally used a broader "skip all net-label-only nets" approach; after rebasing onto current main (post-#530) that regressed 8 examples, so it was reworked into the narrow passive-group predicate above.


workblock100:
Rebased onto current main and reworked after #530 landed: the fix is now a narrow predicate that only affects the exact repro61/79 topology (passive two-pin components joined solely via net labels). All existing snapshots are byte-identical to main, suite is 70/0, and two new repro79 regression tests cover the fix. Ready for review.

workblock100:
Closing since #79 was completed upstream — glad it's fixed. The narrow predicate approach here (skip routing only for fully label-connected passive groups) is available if it's ever useful. Thanks for the reviews of the repo's conventions along the way.

PR #512, opened 2026-06-10
Avoid shifting overlapped traces through schematic component boxes

MustafaMulla29:
`TraceOverlapIssueSolver` used a fixed offset when separating overlapping traces. In this repro, that fixed shift moved the R2 trace into the resistor’s schematic box.

Now, the solver checks the shifted segment against schematic component obstacles. If the default offset would collide, it tries offsets based on the actual obstacle edges and picks the nearest clear one. This keeps the normal shift behavior when it is valid, and only adjusts the offset when it would route through a component.


imrishabh18 (approved):
cc @Sang-it

PR #526, opened 2026-06-11
Prevent the untangle pass from routing traces   through component bodies

MustafaMulla29:
Root cause
The trace arrives at the cleanup stage already valid — routed around C1 at x=-3.7, with one benign C1.1-U1.5 wire crossing:  incoming path:  (-2.7,1.1) → (-3.7,1.1) → (-3.7,-2.2) → (-4.7,-2.2) → (-4.7,-2.0)   ✓ clear of C1

Inside TraceCleanupSolver, the UntangleTraceSubsolver tries to remove that C1.1-U1.5 crossing. But it only evaluates reroute candidates against other traces, never against component bodies — and the only candidate it generates shifts the vertical run to x = -4.9, inside C1's box (x ∈ [-4.965, -4.435]):

untangle reroutes to:  (-2.7,1.1) → (-4.9,1.1) → (-4.9,-2.2) → (-4.7,-2.2) → (-4.7,-2.0)   ✗ through C1

So a stage that's supposed to clean things up takes a valid path and makes it invalid.

The fix
Add a guard in the untangle's candidate evaluation: a reroute candidate is rejected if any of its segments crosses a component, using the existing ChipObstacleSpatialIndex. (The candidate only covers the rerouted corner, never the pin-terminal segments, so the check is clean.)

With the guard, the only available reroute is rejected, so the untangle declines to reroute and keeps the incoming x=-3.7 path as-is — the stage either improves the layout or leaves an already-valid trace untouched, but never breaks it.

This mirrors the guard TraceOverlapShiftSolver already has, so both trace-shifting stages now prevent invalid states at the source rather than relying on a downstream correction pass.


lib/solvers/SchematicTracePipelineSolver/SchematicTracePipelineSolver.ts
```diff
@@ -76,6 +77,7 @@ export class SchematicTracePipelineSolver extends BaseSolver {
   labelMergingSolver?: MergedNetLabelObstacleSolver
   traceLabelOverlapAvoidanceSolver?: TraceLabelOverlapAvoidanceSolver
   traceCleanupSolver?: TraceCleanupSolver
+  traceComponentObstacleAvoidanceSolver?: TraceComponentObstacleAvoidanceSolver
```
imrishabh18: `traceCleanupSolver` from the name mean's it should run at the last and `traceComponentObstacleAvoidanceSolver` should run before.

imrishabh18 (approved):
Take a review from @seveibar before merging

lib/solvers/SchematicTracePipelineSolver/SchematicTracePipelineSolver.ts
MustafaMulla29: traceCleanupSolver is the one that creates the component crossing trace that run it in UntangleTraceSubsolver. The new solver has to run after it.

seveibar: Shouldnt you be fixing/preventing the invalid state trace? I think each stage should strictly be improving things?

seveibar (approved):

seveibar (approved):

PR #538, opened 2026-06-13
fix: resolve long distance pair logic #79

cmalthusian-cyber:
I fixed the issue where the solver was incorrectly drawing traces for net-only connections. Verified locally with the existing test case.
Closed #79


cmalthusian-cyber:
Hi everyone, I’ve updated the configuration to resolve the type check errors, and all CI checks are now passing. I’m ready for your review whenever you have a moment. Thanks

PR #540, opened 2026-06-14
Align close same-net trace segments

Ethan2040:
/claim #34

## Summary

- Add a cleanup pass that aligns close, overlapping same-net trace segments onto the same X/Y coordinate.
- Run that pass after turn minimization and Z-shape balancing in `TraceCleanupSolver`.
- Cover horizontal alignment, vertical alignment, and different-net non-alignment with focused tests.

## Why

Fixes tscircuit/schematic-trace-solver#34. Same-net traces that are already close should visually merge by sharing a coordinate instead of remaining as near-parallel lines.

## Validation

- `bun test`
- `bun run build`
- `bunx biome check --write lib\solvers\TraceCleanupSolver\alignSameNetTraceSegments.ts lib\solvers\TraceCleanupSolver\TraceCleanupSolver.ts tests\functions\align-same-net-trace-segments.test.ts`


Ethan2040:
PayPal works for me as the preferred alternative payout method. I can provide either a PayPal payment link, PayPal.Me link, or PayPal email privately after you confirm that manual PayPal payout is supported.

PR #552, opened 2026-06-17
Fix merge traces

Ami765:
Fixes #34. Collapses collinear trace path segments.


Ami765:
Closes #34. This implementation cleanly identifies and merges overlapping collinear trace paths belonging to the same network.

PR #555, opened 2026-06-19
Add same-net trace segment merge helper

defgtyg:
## Summary

- add a reusable TraceCleanupSolver helper for nearby same-net horizontal/vertical trace segments
- use endpoint segments as stable anchors while only moving interior segments so pin attachments stay pinned
- add regression tests for horizontal, vertical, different-net isolation, and endpoint preservation

## Validation

- GitHub Actions are re-running after removing the broader pipeline hook that changed existing routing snapshots
- Vercel preview is ready
- Not run locally; authored and reviewed through the GitHub web editor in the fork.

/claim #34


defgtyg:
这是来自QQ邮箱的假期自动回复邮件。
 
您好，我最近正在休假中，无法亲自回复您的邮件。我将在假期结束后，尽快给您回复。

yijianli79-tech:
Update after CI: I kept this PR scoped to the reusable same-net segment merge helper plus focused regression tests. A direct TraceCleanupSolver pipeline hook changed several existing routing snapshots, so I backed that out to avoid mixing the reusable geometry fix with broader solver behavior changes. If this direction looks good, I can wire the helper into the cleanup pipeline in a follow-up with whatever collision/snapshot constraints maintainers prefer.

Current checks are passing: format, type check, tests, and Vercel preview.

yijianli79-tech:
Correction: I tested wiring the same-net segment merge helper directly into the final `TraceCleanupSolver` pipeline step, but that broader hook changed six existing routing snapshots. I reverted that hook so this PR stays scoped to the reusable helper plus focused regression coverage while preserving the existing pipeline behavior.

The helper still preserves endpoint-only traces and only moves interior segments. If maintainers want the helper enabled in the cleanup pipeline, I can do that separately with the expected snapshot/collision constraints.

PR #563, opened 2026-06-26
fix: skip MSP pair creation for net-label-only connections (fixes #79)

singhaditya21:
## Problem

When two or more pins are connected only via `netConnections` (net labels like GND, VCC), the `MspConnectionPairSolver` was incorrectly queuing those nets for MSP pair creation and routing physical traces between the pins.

This caused a spurious trace to appear between components that should only display individual net labels, and the `NetLabelPlacementSolver` would then add an extra net label *on that trace* — producing the duplicate label visible in the repro61 screenshot (issue #79).

## Solution

Filter `queuedDcNetIds` in `MspConnectionPairSolver` to exclude any global connectivity net whose component pins do not appear in `directConnMap`. Nets with at least one pin in `directConnections` continue to produce MSP pairs and routed traces as before. Purely label-only nets (only in `netConnections`) receive only net label placements — no redundant trace.

## Changes

- **`MspConnectionPairSolver.ts`**: filter net-label-only nets from the MSP queue
- **New test**: `SchematicTracePipelineSolver_repro61.test.ts` — verifies no MSP pairs and no traces are created for net-label-only connections; GND and VCC labels are still placed
- **Updated test**: `MspConnectionPairSolver_repro1.test.ts` — corrects expected pair count from 4 to 2 (GND was net-label-only, only VCC and EN pairs remain)
- **Updated tests**: `repro130` and `repro126` — updated assertions to reflect that net-label-only nets no longer produce routed traces
- **Regenerated snapshots**: 18 SVG snapshots updated to reflect removal of spurious traces

## Test Results

All 72 tests pass, 4 skipped, 0 failures.

/claim #79
fixes #79


singhaditya21:
Friendly nudge 🙂 — CI is green and this is mergeable whenever you have a moment. Happy to rebase or address any feedback if it helps. Thanks for the great project!

PR #565, opened 2026-06-27
remove noisy console logs

seveibar:


seveibar:
@0hmX surprised it took this long to find it but these were cluttering user logs!!!

PR #570, opened 2026-07-02
add spec for adding textBoxes to the InputProblem

Sang-it:
### What are obstacle?
- Schematic Components have text around them which is not passed to the schematic trace solver, which means the solver doesn't know that they exist, and the traces can go through the texts which make them unreadable.

### Why this?

- We can pass the text bounds as an ```Obstacle``` which will prevent the collisions from happening.

Repro:
Screenshot 2026-06-29 at 11 39 27 PM

### DIFF
--- a/lib/types/InputProblem.ts
+++ b/lib/types/InputProblem.ts
@@ -1,3 +1,4 @@
+import type { Box } from "@tscircuit/math-utils"
 import type { ChipObstacleSpatialIndex } from "lib/data-structures/ChipObstacleSpatialIndex"
 import type { FacingDirection } from "lib/utils/dir"
 
@@ -38,6 +39,7 @@ export interface InputProblem {
   chips: Array<InputChip>
   directConnections: Array<InputDirectConnection>
   netConnections: Array<InputNetConnection>
+  obstacles?: Array<Box>
 
   availableNetLabelOrientations: Record<NetId, FacingDirection[]>
   maxMspPairDistance?: number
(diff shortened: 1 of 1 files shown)


lib/types/InputProblem.ts
```diff
@@ -38,6 +39,7 @@ export interface InputProblem {
   chips: Array<InputChip>
   directConnections: Array<InputDirectConnection>
   netConnections: Array<InputNetConnection>
+  obstacles?: Array<Box>
```
seveibar:
```suggestion
  textBoxes?: Array<Box>
```

lib/types/InputProblem.ts
```diff
@@ -1,3 +1,4 @@
+import type { Box } from "@tscircuit/math-utils"
```
seveibar:
```suggestion
```

lib/types/InputProblem.ts
```diff
@@ -13,6 +13,15 @@ export interface InputPin {
 
   _facingDirection?: "x+" | "x-" | "y+" | "y-"
 }
+
+export interface TextBoxes {
+  chipId: ChipId
+  center: { x: number; y: number }
+  width: number
+  height: number
+  text: string
```
seveibar:
```suggestion
  text?: string
```

lib/types/InputProblem.ts
```diff
@@ -13,6 +13,15 @@ export interface InputPin {
 
   _facingDirection?: "x+" | "x-" | "y+" | "y-"
 }
+
+export interface TextBoxes {
+  chipId: ChipId
```
seveibar:
```suggestion
  chipId?: ChipId
```

seveibar (approved):

PR #580, opened 2026-07-05
Fix: manufacturePartNumber text box overlaps with the net label

imrishabh18:


lib/solvers/SchematicTraceLinesSolver/SchematicTraceSingleLineSolver2/rect.ts
```diff
@@ -1,19 +1,34 @@
...
-  return problem.chips.map(chipToRect)
+export const getObstacleRects = (
+  problem: InputProblem,
+  opts: { textBoxPadding?: RectPadding } = {},
+): ChipWithBounds[] => {
+  const chipRects = problem.chips.map(chipToRect)
+  const textBoxRects = (problem.textBoxes ?? []).map((textBox, index) => {
+    const b = getTextBoxBounds(textBox, opts.textBoxPadding)
+    return {
+      chipId: `text_box_${index}${textBox.chipId ? `_${textBox.chipId}` : ""}`,
```
seveibar: why are you inventing an id here?

lib/solvers/SchematicTraceLinesSolver/SchematicTraceSingleLineSolver2/rect.ts
```diff
@@ -1,19 +1,34 @@
...
 import { getInputChipBounds } from "lib/solvers/GuidelinesSolver/getInputChipBounds"
+import { getTextBoxBounds, type RectPadding } from "lib/utils/textBoxBounds"
 
 export type ChipWithBounds = {
   chipId: string
   minX: number
   minY: number
   maxX: number
   maxY: number
+  isTextBox?: boolean
```
seveibar: a chip is not a text box

seveibar (commented):
why not call this a chip-textbox-collision (you are calling it a chip-collision)? it seems like there's a conflation of chip and text that will cause downstream issues

we could rename the existing one chip-body-collision

PR #584, opened 2026-07-06
Draft: enforce netLabels to follow the `availableNetLabelOrientations`

imrishabh18:
image

image


imrishabh18 (commented):
@Sang-it @seveibar What is the point of `availableNetLabelOrientations`? If you enforce only the available orientations then the snapshots are changing, some don't look good.

I have tried removing the `availableNetLabelOrientations` from the inputProblem cause it seems like all the orientations are available, and even that is turning the snapshots into weird looking placements. I am confused what is this about then from the naming?

Sang-it:
> @Sang-it @seveibar What is the point of `availableNetLabelOrientations`? If you enforce only the available orientations then the snapshots are changing, some don't look good.
> 
> I have tried removing the `availableNetLabelOrientations` from the inputProblem cause it seems like all the orientations are available, and even that is turning the snapshots into weird looking placements. I am confused what is this about then from the naming?

that is more for VCC and GND so that we can pass ```y+``` and ```y-``` for them. If a label cannot be placed (without collisions) in the ```availableNetLabelOrientations``` it just picks a non colliding state to put it in as a fallback.

Rather than removing it might be better to pass all directions for normal netlabel, and ```y+``` and ```y-``` for vcc and gnd.

Right now we only pass ```x-,x+``` for normal net labels.

seveibar:
This pr seems to introduce bugs, you may need to fix the tests. I don’t understand why the bugs are happening- but i think you may find something inside the tests

PR #605, opened 2026-07-06
fix repro51 which is merging the VCC and GND trace, cause the current trace list to avoid was not passed

imrishabh18:


tests/examples/__snapshots__/example49.snap.svg
Sang-it, line 1: Image

overlap

PR #638, opened 2026-07-08
 Fix short schematic trace routing

mohan-bee:
Very short schematic connections between nearby pins could route with unnecessary elbow overshoot, or fail entirely when one endpoint sat slightly inside a neighboring component's obstacle — causing a fallback net label to appear instead of a short trace.

Change

Add a short-distance path case in SchematicTraceSingleLineSolver2, checked before elbow routing (gated by MAX_SHORT_TRACE_DISTANCE).
Normal-distance traces are unaffected — they still go through the existing elbow/collision search path.

BEFORE: 

image

AFTER: 

image


MustafaMulla29 (changes requested):
I dont think we need this(this feels like over-engineering), I feel like its a placement issue. Is that inductor manually placed or auto placed?

seveibar (commented):
Can you moce this logic inside the schematic trace solver? Algorithms dont belong in core- it makes it hard to send reproductions to the algo repo

mohan-bee:
> I dont think we need this(this feels like over-engineering), I feel like its a placement issue. Is that inductor manually placed or auto placed?

inductor is placed manually,, but if feel like having a four unnecessary netlabels that is making a complete mess... , actually schematic-trace-solver is not handling the short distance connection, it fall backs to netlabel, why when two components placed so close having netlabel is won't be good ig, 


@MustafaMulla29

package.json
```diff
@@ -29,5 +29,8 @@
   },
   "peerDependencies": {
     "typescript": "^5"
+  },
+  "dependencies": {
```
MustafaMulla29: Why added tscircuit dep?

mohan-bee: ummm sorryy accidentally added reverted bad committtt sorry

mohan-bee: updated ! sorry !

seveibar (changes requested):
1. Dont clutter classes with instance methods that dont need to be instance methods (no state or primarily not state mutating)
2. Proof that this works? I dont see it in the snapshot working

mohan-bee:
@seveibar  

before with netlabels

image

after
image

seveibar:
How is that fixed?

seveibar:
@mohan-bee ok i think i understand- just refactor the private methods into separate files (do not do this antipattern with instance methods!!)

seveibar (approved):

MustafaMulla29 (approved):

PR #645, opened 2026-07-08
 Respect provided pin facingDirection when correcting pins inside expanded chip boxes

MustafaMulla29:
## What

`correctPinsInsideChips` moves any pin that sits *inside* its chip's bounding box out onto the nearest edge, so the router has a perimeter point to start from. This changes that step to honor an explicit `_facingDirection`: when a pin already declares which way it faces, snap it to the edge **along that direction** and keep the facing, instead of snapping to the geometrically nearest edge and clearing it.
Nearest-edge snapping is still the fallback for pins with no `_facingDirection`, so existing behavior is unchanged.

## Why

Consumers (e.g. `@tscircuit/core`) hand the solver a **text-inclusive** chip box — the component's box grown to enclose its reference-designator/value text. For a *small-variant* part with a *large* ref designator (a resistor labeled `R_FAULT_PULLUP`, a decoupling cap, …), that box is much wider/taller than the part, so the component's two ports end up **inside** the box rather than on an edge.

`correctPinsInsideChips` then moved each port to the *nearest* edge. For a horizontal resistor whose wide ref text pushed the left/right edges far away, the nearest edge is the **bottom** — so both terminals were snapped downward and
re-detected as facing `y-`. The result: connections/net labels dropped *below* the part instead of leaving from its left/right sides.

The port's true facing is already known from the schematic symbol, and the solver already respects `pin._facingDirection` everywhere else (`pin._facingDirection ?? getPinDirection(...)`). The one place that ignored it — and actively cleared it — was this correction step. This makes it consistent: trust the declared facing.

This is the solver half of the fix; the consumer supplies `_facingDirection` from the symbol (`@tscircuit/core` PR: https://github.com/tscircuit/core/pull/2614). Neither half works alone — sending the facing has no effect while this step overwrites it.

## Before / after

- Before: small resistor with a long ref designator → both pins face `y-`, trace/net labels route below the part.
- After: pins keep `x-`/`x+`, connections leave from the sides.

Before:
Screenshot 2026-07-08 at 4 06 41 PM

After:
Screenshot 2026-07-08 at 4 07 05 PM

### DIFF
--- a/lib/solvers/SchematicTracePipelineSolver/correctPinsInsideChip.ts
+++ b/lib/solvers/SchematicTracePipelineSolver/correctPinsInsideChip.ts
@@ -13,6 +13,25 @@ export const correctPinsInsideChips = (problem: InputProblem) => {
 
       if (!isInside) continue
 
+      // When the pin already knows which way it faces (e.g. the consumer set it
+      // from the schematic symbol), snap it to the edge along that direction and
+      // keep the facing. A pin can end up inside the box when the box was
+      // expanded to include the component's reference-designator text; snapping
+      // to the nearest edge would move it to the wrong side (e.g. a horizontal
+      // resistor whose wide ref text makes the bottom edge the closest one).
+      if (pin._facingDirection) {
+        if (pin._facingDirection === "x-") {
+          pin.x = bounds.minX
+        } else if (pin._facingDirection === "x+") {
+          pin.x = bounds.maxX
+        } else if (pin._facingDirection === "y-") {
+          pin.y = bounds.minY
+        } else {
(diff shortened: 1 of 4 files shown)


imrishabh18 (approved):
cc @seveibar @Sang-it

seveibar (approved):

seveibar (commented):
Why does _facingDirection have an underscore 🤔 module boundary should hide this

MustafaMulla29:
> Why does _facingDirection have an underscore 🤔 module boundary should hide this

Its the original API, happy to rename it if you want

PR #649, opened 2026-07-09
Give each pushed trace its own corridor when escaping a shared net-label block

### LINKED ISSUE #597: JSON Bug Report:

[_board#756 _-SchematicTracePipelineSolver-solver-input.json](https://github.com/user-attachments/files/29721099/_board.756._-SchematicTracePipelineSolver-solver-input.json)

### Notes

_No response_

abdalraof-albarbar:
Fixes #597.

When two different-net traces both collide with the **same** net-label block, `generateHorizontalSegmentPushCandidate` pushed each colliding vertical segment to `bounds.maxX + LABEL_CLEARANCE` — the *identical* corridor — so the two traces ended up drawn stacked on one line (a 2.6-unit collinear overlap right of `J2` at x≈15.7 in bug-report-20260706T213649Z).

### Fix
After computing the push X, slide the corridor outward past any **different-net** trace already routed there, so each escaping trace gets its own parallel corridor. Scoped to different nets only (same-net stacking is intentional / handled elsewhere), and bounded to `MAX_CORRIDOR_SHIFTS`.

### Before / after (traces only, J2 corridor)
corridor before/after

### Verification
- Different-net trace overlaps in bug-report-20260706T213649Z drop from **4 → 2** — both vertical-corridor overlaps (x≈15.7 and x≈-10.65) are resolved. (The remaining 2 are unrelated *horizontal* overlaps near x≈17.7, a separate defect out of scope here.)
- Only that one bug-report snapshot changes; the rest of the suite is untouched (0 regressions across 106 tests). Regenerated the snapshot.


seveibar (approved):
good, maybe not perfect because it seems like there is still some trace crowding

Issue #620, opened 2026-07-07
JSON Bug Report:


ShiboSoftwareDev:
### Solver input JSON attachment

[_board#258 _-SchematicTracePipelineSolver-solver-input.json](https://github.com/user-attachments/files/29748279/_board.258._-SchematicTracePipelineSolver-solver-input.json)

tscircuitbot:
Created or updated https://github.com/tscircuit/schematic-trace-solver/pull/621 for this JSON bug report.

Report id: `bug-report-20260707T134549Z`

PR #650, opened 2026-07-09
Never route net-label direct connections as wires (LongDistancePairSolver + MspConnectionPairSolver)

### LINKED ISSUE #643: JSON Bug Report:

[_board#91 _-SchematicTracePipelineSolver-solver-input.json](https://github.com/user-attachments/files/29795016/_board.91._-SchematicTracePipelineSolver-solver-input.json)

### Notes

_No response_

abdalraof-albarbar:
Closes #643

### Problem

A direct connection that carries a `netLabelWidth` is a **net label** — the caller intends `NetLabelPlacementSolver` to render it as a label attached to the pin, not as a wire drawn across the schematic. Two independent pair generators turned these connections into routed traces:

1. **`LongDistancePairSolver`** paired every in-net unconnected pin with its nearest neighbors as a full-wire candidate without checking `netLabelWidth`, so far-apart net-label connections were drawn as spurious traces.
2. **`MspConnectionPairSolver`** never checked `netLabelWidth` either. A net-label connection whose two pins are within `maxMspPairDistance` becomes a primary MSP pair and is routed as a wire by `SchematicTraceLinesSolver` — *before* `LongDistancePairSolver` runs — drawing a wire **alongside** the net label.

The open bug reports only exercised (1): all of their net-label connections happen to be farther apart than `maxMspPairDistance` (11–15 vs 2.4), so they all went through the long-distance path. The close-distance case (2) was a latent gap.

### Fix

Extract the pair-key computation into a shared helper so the two pair generators can't drift:

```ts
export const getNetLabelDirectConnectionPairKeys = (
  inputProblem: InputProblem,
): Set<string> => {
  const pairKeys = new Set<string>()
  for (const directConnection of inputProblem.directConnections) {
    if (directConnection.netLabelWidth == null) continue
    pairKeys.add([...directConnection.pinIds].sort().join("--"))
  }
  return pairKeys
}
```

- `LongDistancePairSolver` uses it to skip net-label pairs during candidate generation (as before, now via the shared helper).
- `MspConnectionPairSolver` uses it to skip net-label pin pairs in both the two-pin and MSP-tree branches, so a close net-label connection is never routed as a wire.

The exclusion targets the exact net-label pin pair (a normal connection between the same-distance pins is unaffected).

### Before / after

`bug-report-20260708T095725Z` — the two long horizontal traces from the left-hand pins were net-label connections (`U1.14-RTEMP.1`, `U1.13-RFSET.1`). They are gone after the fix; the pins render as labels.

before/after

### Impact

Spurious net-label wires (traces that touch a `netLabelWidth` pin) in the existing bug reports, before → after:

| bug report | before | after |
| --- | --- | --- |
| `bug-report-20260708T095725Z` (#643) | 2 | **0** |
| `bug-report-20260707T140410Z` (#626) | 1 | **0** |
| `bug-report-20260707T134549Z` (#620) | 5 | **4** |

Only these three snapshots change. The `MspConnectionPairSolver` change is additive robustness — it changes **no** existing snapshot (all bug-report net-label connections are long-distance), and is covered by a new focused regression test asserting a close net-label connection is skipped while a close normal connection still pairs.

`bun test` 108/108 pass · `tsc` clean · `biome format` clean.


seveibar (changes requested):
based on the PR description, this does not seem like desirable behavior

sometimes VCC is routed with a wire and net label, that is very common- so i don't know why we would disable that

ALso note, your usage of "direct" in the PR title is EXTREMELY confusing- a direct connection is NOT a net connection, indirect and net connection are essentially synonymous

PR #654, opened 2026-07-10
Snap power/ground net labels to a pin-aligned corner

MustafaMulla29:
Power/ground rail net labels (`VCC`, `GND`, `V3_3`, …) were placed at arbitrary points along their trace — typically at a mid-trace bend — instead of sitting as a clean stub aligned to a pin. In repro46, `V3_3` floated at a notch in the wire
between C2 and U1, and `GND` looped awkwardly below, neither aligned to a pin.

Two causes in `VccNetLabelCornerPlacementSolver` (the solver that snaps rail abels to trace corners):

1. It only processed labels where `netId === "VCC"`, so `GND`, `V3_3`, and every  other named rail were never considered.
2. For the labels it did process, it snapped to the *closest* corner — which for a two-pin rail net is usually the mid-trace bend, not a pin-aligned position.

## Fix

1. **Process all vertical rail labels.** Detect rails by their fixed vertical orientation (`label.orientation === "y+" || "y-"`) rather than a `netId` string match. Rails point up/down; signal labels (`x+`/`x-`) point sideways and are left alone.
2. **Only snap to pin-aligned corners.** A corner is a candidate only when it  lines up with one of the trace's pins along the label's stub axis (same x for a vertical label), so the label reads as a clean stub straight off that pin. Non-aligned corners aren't offered, which also leaves already-well-placed labels untouched.

Screenshot 2026-07-10 at 4 03 58 PM

Screenshot 2026-07-10 at 4 05 22 PM

Motivation:
Screenshot 2026-07-10 at 4 09 10 PM


MustafaMulla29:
Currently the file named `VccNetLabelCornerPlacementSolver` we should rename it with better name, did not do it in this pr coz the change wont be visible. Should i do it in this pr only?

imrishabh18 (approved):

PR #658, opened 2026-07-11
fix: handle empty or undefined connections arrays without crashing

### LINKED ISSUE #657: [BUG] Router crashes on empty connections array
**Describe the bug**
   All solvers crash when `connections: []` is passed. Need early return guard.
   
   **Files affected**
   - CapacityMeshSolver.ts
   - DirectLineSolver.ts 
   - etc.
   
   **Proposed fix**
   Add `if (!input.connections?.length) return { traces: [] }` to each solver.
   
   **Would you accept a PR for this? Happy to add a bounty label if you want.**

holistis:
## Summary

Fixes #657. Solvers threw `TypeError: ... is not iterable` or `Error: Unexpected numItems value: 0` when `chips`, `directConnections`, or `netConnections` were empty or `undefined`.

- Normalize `chips`, `directConnections`, and `netConnections` to `[]` in `cloneAndCorrectInputProblem` â€” the single pipeline entry point, so all downstream solvers receive valid arrays
- Add `?? []` fallbacks in `getConnectivityMapsFromInputProblem` as defense-in-depth (this function can also be called independently)
- Guard `ChipObstacleSpatialIndex` against `chips.length === 0` â€” Flatbush throws when constructed with 0 items; skip index construction and return early
- Type `spatialIndex` as `Flatbush | null` (was `Flatbush`) and guard `getChipsInBounds` against `null`

## Test plan

- [ ] `bun test tests/repros/repro-empty-connections.test.ts` â€” 3 new regression tests (empty arrays, undefined connections, all fields undefined)
- [ ] `bun test tests/examples/` â€” 50 existing example tests still pass
- [ ] `bun test tests/repros/` â€” all existing repro tests still pass
- [ ] `bunx tsc --noEmit` â€” no type errors

### DIFF
--- a/lib/data-structures/ChipObstacleSpatialIndex.ts
+++ b/lib/data-structures/ChipObstacleSpatialIndex.ts
@@ -10,7 +10,7 @@ export interface SpatiallyIndexedChip extends InputChip {
 
 export class ChipObstacleSpatialIndex {
   chips: Array<SpatiallyIndexedChip>
-  spatialIndex: Flatbush
+  spatialIndex: Flatbush | null
   spatialIndexIdToChip: Map<number, SpatiallyIndexedChip>
 
   constructor(chips: InputChip[]) {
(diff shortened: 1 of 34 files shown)


holistis:
Hi, just checking in on this one, it has been about three weeks. I noticed #659 is also open on the same issue, happy to close mine if you would rather go with that one, or let me know if you want any changes here.

PR #663, opened 2026-07-14
Align same-net schematic rails during trace cleanup

MustafaMulla29:
## Summary

- reuse `TraceCleanupSolver` for a late same-net rail-alignment cleanup pass instead of introducing a standalone solver
- align horizontal and vertical rails only when they belong to the same global net and the same side of the same component
- derive movable trace IDs from the earlier cleanup output, excluding generated net-label connectors without adding shared trace-role metadata
- compare every candidate against the unchanged route using rendered visible length, turn count, logical length, crossings, and displacement
- keep `alignSameNetRails.ts` as a 66-line orchestration entry point, with discovery, grouping, geometry, scoring, movement, and label-anchor checks in focused helpers
- add focused vertical/horizontal repros and safety regressions, with exactly one test per test file

## Root cause

Related traces are routed independently, so collision avoidance can leave several staggered, parallel runs around one component even when they represent one continuous same-net rail.

The earlier implementation grouped nearby same-net geometry too broadly. It also introduced a dedicated solver and a `traceRole` field solely to distinguish routed traces from generated label connectors. That allowed unrelated routes to move and made snapshot changes difficult to justify.

## Implementation

Rail alignment now runs as an optional late operation of the existing `TraceCleanupSolver`, after trace/net-label collision handling and before the existing final `NetLabelNetLabelCollisionSolver`.

The cleanup pass:

- considers only positive-length internal orthogonal rails; moving only their internal rail points preserves trace endpoints and orthogonality by construction
- associates each rail with an endpoint component side using the existing pin-facing-direction helper
- groups only same-global-net rails on the same component side, connected by shared-pin topology or overlapping spans
- evaluates only coordinates already present in the group
- treats traces present before label-connector generation as movable and all later traces as immutable
- rejects candidates that introduce component/text collisions, label overlap, other-net coincidence, immutable-trace coincidence, extra crossings, or broken label anchors
- includes the unchanged geometry as the baseline and accepts alignment only when rail count decreases without worsening turns and rendered readability

The implementation reuses the existing pin-direction, turn-count, path-simplification, collision, and point-on-trace utilities. No component names, pin numbers, net names, or RP2040 coordinates are special-cased.

## Impact and blast radius

The RP2040 gamepad reproduction now forms one readable GND rail on each side. New generic fixtures prove the same behavior for vertical and horizontal component sides.

Every changed solver snapshot was visually reviewed. Unrelated changes from the broad implementation were reverted. The retained changes either shorten a route, consolidate equal-length bends, reduce turns, or replace duplicate visible geometry with a shared rail; none add crossings or obstacle/label collisions.

The exact `U3` / `V3_3` regression has a pipeline test proving the real pin-to-pin loop remains local and is not stretched toward its generated label connector.

## Validation

- solver focused alignment/repro tests: 11 passed across 11 files
- solver full suite: 114 passed, 4 skipped, 0 failed
- solver `bunx tsc --noEmit`
- solver `bun run build`
- solver `bun run format`
- solver `git diff --check`
- headless pipeline artifacts reviewed for the vertical, horizontal, and RP2040 repros
- bun-linked core targeted tests:
  - `group-match-adapt3`
  - RP2040 gamepad trace-alignment repro
- bun-linked core full suite: 1,167 passed, 31 skipped, 0 failed


MustafaMulla29:
@seveibar is this the correct pattern? Or it should be a separate solver?

lib/solvers/SchematicTracePipelineSolver/SchematicTracePipelineSolver.ts
```diff
@@ -303,6 +305,48 @@ export class SchematicTracePipelineSolver extends BaseSolver {
         },
       ],
     ),
+    definePipelineStep(
+      "finalTraceCleanupSolver",
```
seveibar: oof, NEVER EVER USE THE WORD `final`!!!!!!

seveibar: you can enumerate e.g. `traceCleanupSolver2` or something like that but NOTHING is ever final!

seveibar (approved):
see comment

PR #664, opened 2026-07-15
Prevent non-finite hues for long color identifiers

### LINKED ISSUE #651: getColorFromString returns hsl(NaN, ...) for very long strings

```ts
```

partyplatter08-lab:
## Summary

- preserve existing `getColorFromString` results for ordinary identifiers
- switch to modulo-360 accumulation before the hash can overflow, keeping arbitrary-length identifiers finite and deterministic
- verify the reported long-string case and ensure long suffixes still affect the generated color

## Testing

- `bun test`
- `bun run build`
- `bun run format:check`
- `bunx tsc --noEmit`

Closes #651

/claim #651

### DIFF
--- a/lib/utils/getColorFromString.ts
+++ b/lib/utils/getColorFromString.ts
@@ -1,7 +1,14 @@
 export const getColorFromString = (string: string, alpha = 1) => {
   // pseudo random number from string
-  const hash = string.split("").reduce((acc, char) => {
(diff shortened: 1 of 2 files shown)


partyplatter08-lab:
Friendly ping — this PR remains open and CI-clean on our side (tracked bounty ~$50).

Happy to address any review comments or adjust scope if something no longer fits the issue.
No rush; just keeping the thread current.

PR #668, opened 2026-07-16
Fix U-shaped endpoint obstacle detours

MustafaMulla29:
## Summary

- extend the existing endpoint-collision detour generator from L-shaped base elbows to fixed-pair U-shaped base elbows
- preserve the untouched remainder of the U path so the required endpoint-facing approach is not lost
- keep optional long-distance routing and endpoint-chip routing on their established strategies because those require net-level or multi-trace selection
- update repro47 and the equivalent example11 snapshots from label fallback to collision-free traces

## Root cause

Same-facing pins produce a four-point U-shaped base elbow. When an intermediate component blocked the first or last segment, the solver used its generic adjacent-segment shift. That shift cannot insert the extra dogleg needed to go around the obstacle while preserving the U bend and the pin-facing approach, so the candidate search exhausted and the pair fell back to net labels. The existing endpoint-detour helper already models the correct operation, but only retained three-point L paths.

This change generalizes that helper to retain the suffix/prefix of four-point U paths. It does not add circuit-specific coordinates, distance exceptions, or iteration increases.

Repro introduced in #667.

## Validation

- `bun test` — 122 pass, 4 skip, 0 fail
- `bun run build`
- `bun run format:check`
- generated and inspected every pipeline stage for repro47
- reviewed every changed snapshot; final blast radius is only repro47 and equivalent example11, and both replace the same two failed pairs with real obstacle-clearing traces
- explicitly rejected broader variants that regressed example40 or reduced real traces in bug-report-20260706T220324Z


seveibar (approved):
To be clear, the trace snapshots here are not optimal and are actually kind of confusing-

PR #672, opened 2026-07-16
fix: route endpoint U-detours without trace overlaps

MustafaMulla29:
## Summary

- prefer exterior four-point U detours whose internal rails stay outside the endpoint pin band
- resolve internal collinear point contacts in the existing trace-overlap stage
- shift adjacent overlap segments atomically from their original axes
- compare mirrored separation candidates only for simple pairwise overlaps
- preserve deterministic existing behavior for compound overlaps
- add focused repro47, example16, and overlap-solver regression coverage

## Root cause

Two geometry issues combined in the failing layouts:

1. Four-point endpoint U detours inherited the L-detour length-first ordering, so an interior U could win before the clearer exterior route.
2. The overlap solver only detected positive-length collinear overlaps. Different-net internal rails that met at a single endpoint were therefore left as false junctions. It also mutated adjacent participating segments sequentially, so a shared point could be moved twice or a later segment could temporarily appear diagonal.

Applying candidate scoring indiscriminately to compound overlaps changed established rail ordering and affected downstream net-label placement. Filtering or rotating net-label candidates would address that symptom in the wrong solver.

## Fix

Endpoint detour ranking remains in the route generator. Four-point U candidates compare pin-band clearance before path length; existing L-detour ordering is unchanged.

Trace contact resolution remains in `TraceOverlapShiftSolver`:

- positive-length overlaps and internal point contacts are represented as distinct interaction kinds
- segment axes are captured before mutation, and each shared coordinate is shifted once
- point contacts try moving one island in either direction
- simple pairwise overlaps compare the two mirrored, obstacle-aware separations
- compound overlaps keep the solver's established deterministic separation because one candidate score cannot attribute downstream intersections to an individual participating rail
- candidates use the normal different-net intersection count with no terminal-segment exclusions

There are no changes to `AvailableNetOrientationSolver`. Requested power, ground, and ordinary net-label orientations continue through the existing placement flow.

## Impact

- repro47 uses the exterior endpoint detours and keeps both relevant different-net trace pairs separated
- example16 retains the required `y-` GND label orientation with zero different-net intersections
- the reported `bug-report-20260716T144856Z` case passes
- no circuit-specific coordinates, constructor filtering, snapshot exceptions, iteration-limit changes, or terminal-crossing score exceptions

## Validation

- `bun test` — 128 pass, 4 skip, 0 fail
- `bun run build`
- `bun run format:check`
- `git diff --check`
- linked local core test: `schematic-section-autolayout-no-section.test.tsx` — 1 pass, 56 assertions


MustafaMulla29 (commented):
There are still some trace overlaps in obvious clean places, need to be fixed in next prs

tests/repros/__snapshots__/repro47-endpoint-obstacle-detour.snap.svg
MustafaMulla29, line 1: Such as the middle portion traces

seveibar (approved):

PR #673, opened 2026-07-16
Use net labels for long connections

mohan-bee:
motivation: in tscircuit there were some traces drawn for too long which is little hard to track , ig we can replace those with netlabels to keep that clean. 

image

ref: 

image
 
 
this fixes : 

Prevents long named-net connections from being routed across the schematic by using net labels instead, while keeping short traces and explicit direct connections unchanged.


MustafaMulla29 (changes requested):
I dont think this is needed, the more traces we solve is the better! Maybe ask seve about this.

mohan-bee:
> I dont think this is needed, the more traces we solve is the better! Maybe ask seve about this.

but when i try to refer some schematics online , it looks , like having netlabes instead of long traces < 

image

MustafaMulla29:
Maybe take suggestion from seve

PR #681, opened 2026-07-17
remove unnecessary labels after routing

mohan-bee:
motivation: traces routed successfully but still seeing some unnecessary labels


mohan-bee:
these labels is necessary for readability .

PR #683, opened 2026-07-17
Fix close opposing-pin routes with collision-aware adaptive elbows

mohan-bee:
Motivation

Closely placed components could fail schematic routing and incorrectly fall back to net labels because the default elbow overshot into the endpoint components.

Change

When the default elbow collides, the solver now tries a distance-scaled elbow and uses it only when collision-free. Existing valid routes remain unchanged.

Adds regression coverage for bug-report-20260708T053736Z, including assertions that the trace is routed without net-label fallback.


lib/solvers/SchematicTraceLinesSolver/SchematicTraceSingleLineSolver2/SchematicTraceSingleLineSolver2.ts
```diff
@@ -28,6 +28,29 @@ import { getObstacleRects, type ObstacleRect } from "./rect"
 
 type PathKey = string
 
+const calculateElbowWithOvershoot = ({
```
MustafaMulla29: Maybe give it a better name, overshoot doesn't look good

MustafaMulla29 (approved):

PR #685, opened 2026-07-19
Prevent explicit netlabels from expanding into long schematic traces

rushabhcodes:
## Summary

- add `preExistingNetLabelPinIds` to the schematic trace solver input
- prevent the long-distance fallback from routing across endpoints that already have user-placed netlabels
- skip duplicate automatic label placement for pins and routed islands already represented by explicit netlabels
- preserve existing long-distance routing for connections without explicit labels

## Root cause

Normal MSP routing correctly rejects these connections because they exceed `maxMspPairDistance`. The long-distance fallback then reconnects the same endpoints with full-width schematic traces because it has no knowledge that one side is already represented by an explicit netlabel. Automatic netlabels are added later, but the unnecessary routes remain.

The new input field carries that intent into the solver. Long-distance candidates touching an explicitly labeled pin remain separate net islands, allowing the remaining endpoints to terminate at automatic netlabels without drawing wires across the schematic.

## Cross-repository reproduction

The user-facing regression is reproduced in [tscircuit/core#2725](https://github.com/tscircuit/core/pull/2725).

This PR also includes a minimal solver-level regression test and verifies that unlabeled endpoints still retain clear long-distance routes.

## Before and after

| Before | After |
| --- | --- |
| Before: full-width traces are routed between explicitly labeled nets | After: endpoints terminate at netlabels without full-width traces |

The dashed lines are connectivity guides in the solver visualization. The solid full-width routes visible before the fix are no longer emitted afterward.

## Validation

- `bun test tests/repros/repro152-explicit-netlabels-long-traces.test.ts tests/solvers/SchematicTracePipelineSolver tests/solvers/MspConnectionPairSolver tests/solvers/NetLabelPlacementSolver`
- `bunx tsc --noEmit`
- normalized final-routing snapshot verification


MustafaMulla29 (commented):
The repro you merged was with manual netlabels and trace solver doesn't know about manual netlabels. I think auto netlabels won't have the same outcome

PR #690, opened 2026-07-22
fix: resolve stuck net-label collisions with a least-overlap fallback

### LINKED ISSUE #687: JSON Bug Report:

[_board#157553 _-SchematicTracePipelineSolver-solver-input.json](https://github.com/user-attachments/files/30245182/_board.157553._-SchematicTracePipelineSolver-solver-input.json)

### Notes

_No response_

Hero988:
Fixes #687. Reproduction: #691 (merged — `tests/repros/repro-netlabel-collision-687` pins the buggy stacked state this PR flips).

## Problem

In the user-submitted board `bug-report-20260721T221026Z` (#688), the `V1V1` net label is left stacked directly on top of the `QD3` and `QCLK` port labels. `NetLabelNetLabelCollisionSolver` only resolves a colliding pair if some candidate is *fully* collision-free; when both labels exhaust their candidates the pair is added to `skippedCollisionKeys` and the worst-case overlap ships unchanged. #691 minimizes this to a 5-chip scene with the same shape and net names.

## Fix

A least-bad fallback, used **only** when the strict search fails (cases that resolve today are untouched):

- Each failing candidate gets a penalty: overlap area with other labels **plus trace-crossing length weighted so heavily (×10) that a wire through a label's text always outweighs any plausible label-label overlap saving**. Only the candidate's host segment (the one its anchor sits on) is exempt. Chip/text collisions stay disqualifying.
- When both labels' queues are exhausted, the best candidate is applied **iff it strictly improves on the current penalty**; the pair is then marked skipped as before. Strict improvement makes total penalty monotonically decrease, so fallback moves cannot cycle.

An earlier revision of this PR weighted own-net crossings at only 0.05 and exempted *any* own-net segment touching the anchor — at a trace corner both adjacent segments qualified, so the fallback could park the label on top of its own wire (review caught this). Now the crossing dominates and the exemption is host-segment-only.

## Result

- **Minimal repro (#691 scene):** `V1V1` leaves the QD3/QCLK stack and tucks beside its own vertical wire. V1V1×QD3 and V1V1×QCLK go to **0**, total overlap 0.162 → 0.119 (the scene has no fully-free spot by construction), and a new assertion pins **zero trace-crossing length through the label**.
- **Original board:** honestly unchanged. That region is congested at wire pitch — of V1V1's 50 candidates, every one except its current spot crosses ≥ 0.42 of other wires — so the fallback correctly refuses to trade a label-label overlap for a label-on-wire. Its snapshot has zero diff vs main. Fixing that board's case needs a bigger action space (moving the port-label column, rerouting, or off-trace labels with a leader) — see PR discussion.

## Tests

`bun test`: 144 pass / 0 fail; `bunx tsc --noEmit` clean; `biome format` clean.

## Note / possible follow-up

`netLabelWidthOf`/`netLabelHeightOf` un-rotate label dims assuming the rotated-box model from `geometry.ts`, but upstream placements use horizontal-text dims (a `y+` label arrives as 0.6×0.42, so `x±` candidates get 0.42×0.6). Not blocking, left alone to avoid snapshot churn.


seveibar (changes requested):
we never merge in this repo without a clear reproduction, start with a reproduction PR

Hero988:
Repro PR opened as requested: #691 — reproduction from the user-submitted board bug-report-20260721T221026Z (the same input this PR's red→green test uses), asserting the current stuck-overlap behavior with a visual snapshot. Once #691 is in I'll rebase this fix on it and flip the repro assertions to green.

imrishabh18 (changes requested):
Where is the before and after snapshot?

Hero988:
Here are the exact before/after snapshots:

| Before (current main) | After (this PR) |
|---|---|
| Before: V1V1 overlapping QD3 and QCLK | After: overlap resolved |

The changed SVG in this PR is the after snapshot; the reproduction PR #691 separately pins the buggy state.

seveibar (commented):
it's not really clear to me if this fixes something without first a PR to make a repro

Hero988:
@seveibar the reproduction PR already exists — it's #691, opened right after your Jul 22 review. It contains only the repro: the user-submitted board `bug-report-20260721T221026Z` (from #688), a test asserting the current buggy behavior (V1V1 stacked on QD3/QCLK, overlap > 0), and a snapshot of the pile-up — no fix code.

Proposed order: merge #691 first so CI pins the bug on a real board, then this PR rebases onto it and flips exactly those assertions to `.toBe(0)`.

Happy to restructure if you'd prefer a different shape (e.g. folding the repro commit into this PR as its first commit).

Hero988:
Restructured so the before/after is clearly visible: this PR is now rebased on the simplified repro from #691, and the fix commit flips that repro's snapshot and assertions in its own diff.

| Before (main — solver gives up) | After (this PR — least-overlap fallback) |
|---|---|
| Before: V1V1 stacked on QD3/QCLK | After: V1V1 clear of the column |

Measured on the minimal scene: `V1V1`×`QD3` and `V1V1`×`QCLK` overlap drop from 0.052 / 0.110 to **0**; total label-label overlap 0.162 → 0.025 (the residual is inherent — the scene has no fully-free spot, which is what triggers the bug). The full-board test is gone; the original board stays covered by its existing bug-report snapshot, which this PR's diff also shows improving.

Once #691 merges this PR reduces to just the solver change + the snapshot flip.

seveibar:
@Hero988 what does the actual schematic reproduction look like? Not inside this library, but the actual schematic? Just trying to see what this case is about- the solution looks wrong- your new net label position overlaps the trace and is flipped

Hero988:
Sorry about that, looks like you were right: the previous commit's placement was wrong. Full picture below.

**What this case actually is:** #687 is an auto-filed JSON bug report from a user board (`board#157553`) — it only contains the solver-input JSON, no original circuit source or screenshot. So to show the actual schematic I rendered the scene schematic-style, to scale, from the exact geometry (chip outlines, pins, wires, label boxes at their true coordinates and sizes). This is the merged #691 repro scene, which is the real board's problem region minimized with the same shape and net names: `U1`'s pin column is walled in by the QD0–QCLK port-label column, C1/C2 above and below, and U3 closing the corridor, while the `V1V1` wire (pin 1 → pin 7) wraps around it. There is **no fully collision-free spot for the V1V1 label anywhere on its wire** — that's the bug: the solver gave up and stacked it on QD3/QCLK.

before — main

**What you saw (previous commit):** the fallback parked V1V1 at the wire's top corner, with the vertical wire running straight through the label, flipped downward. Root cause: the penalty exempted *any* own-net segment touching the anchor point — at a trace corner **both** adjacent segments pass that test, so the wire through the label's body cost zero.

rejected — previous commit

**Now (95c5635):** only the candidate's host segment is exempt, and trace crossings dominate the penalty (×10 weight — a wire through label text always outweighs any label-label overlap saving). V1V1 tucks beside its own vertical wire: V1V1×QD3 and V1V1×QCLK go to 0, total overlap 0.162 → 0.119 (some residual is inherent — the scene has no free spot by construction), and a new test assertion pins **zero trace-crossing length through the label**.

new — corrected fallback

**Honest limit on the original board:** with wire-crossings ruled out, the original board's region is congested at wire pitch — of V1V1's 50 candidates there, every single one except its current spot crosses ≥ 0.42 of other wires. So the corrected fallback leaves that board **unchanged** (still stacked, zero snapshot diff vs main) rather than putting the label on a wire. Actually fixing that board's case needs a bigger action space than "slide this label along its own trace": moving/reflowing the port-label column, rerouting the V1V1 trace out of the pin fan-out, or allowing an off-trace label with a leader line.

So: this PR now = a conservative fallback that resolves stuck stacks **where a wire-clear improvement exists** and never makes anything worse. Happy to land it as that, extend it in whichever of the directions above you prefer, or close it if you'd rather attack the congested case with a different mechanism.

I underestimated this fix, my apoglies for this

seveibar:
@Hero988 make sure you have a real issue when creating issues against algos, we don't want to create synthetic examples

Hero988:
Understood. I learnt a lot here, so thank you :)

PR #691, opened 2026-07-22
test: repro #687 — V1V1 net label left stacked on QD3/QCLK port labels

Hero988:
Reproduction-only PR for #687, as requested in the review on #690.

Input is a minimal 5-chip scene (`tests/repros/assets/repro-netlabel-collision-687.input.json`) minimized from the user-submitted board `bug-report-20260721T221026Z` (#688), preserving its exact failure mechanism and net names.

**What it shows:** the `V1V1` trace label has no strictly collision-free candidate anywhere along its trace — the pin column is walled in by the QD0–QD3/QCLK port-label column, chips above/below, and a chip closing the corridor — so `NetLabelNetLabelCollisionSolver` gives up and leaves it stacked on the `QD3` and `QCLK` port labels (total stacked overlap ≈ 0.16; the original board shows ≈ 0.168 in the same shape).

| Current behavior (pinned by this PR) |
|---|
| V1V1 stacked on QD3/QCLK |

- `tests/repros/repro-netlabel-collision-687.test.ts` asserts the CURRENT buggy behavior (overlaps > 0) so CI pins the bug; a fix should flip those two assertions to `.toBe(0)`.
- #690 is rebased on this branch and flips exactly this snapshot, so its diff shows the before/after.


Hero988:
The repro is now a minimal 5-chip scene instead of the full user board (snapshot went from 345 lines to 80).

The scene reproduces the exact mechanism from bug-report-20260721T221026Z: a walled-in pin column where the `V1V1` trace label has no strictly collision-free candidate, so `NetLabelNetLabelCollisionSolver` gives up and leaves it stacked on the `QD3`/`QCLK` port labels (overlap ≈ 0.16, same as the original board's ≈ 0.168).

| Current behavior (this PR pins it) |
|---|
| V1V1 stacked on QD3/QCLK |

The red `V1V1` label sitting on the port-label column (bottom-left) is the bug. #690 is rebased on top of this and flips exactly this snapshot, so the before/after is visible in its diff.

seveibar (approved):

PR #696, opened 2026-07-23
Prevent schematic traces from crossing component text

techmannih:
## Summary

- prevent schematic traces from crossing the raw bounds of component text
- keep endpoint label padding relaxations without allowing traces through the text itself
- add local internal-segment detours for long-distance routes blocked by endpoint text
- preserve the existing `U1.14-RTEMP.1` upper trace instead of replacing it with fallback net labels
- add semantic regression assertions and update the affected snapshots

## Context

The TPS61222 repro from tscircuit/core#2785 routed GND through the manufacturer part number. Treating raw text bounds as hard obstacles fixed that collision, but initially caused an existing long-distance trace to disappear because shifting its entire rail overlapped another trace.

Long-distance routes now detour locally around endpoint text while preserving their endpoint legs. This keeps the upper trace present and avoids both the MPN and reference-designator text.

## Testing

- 139 tests passed, 4 skipped, 0 failed
- `bunx tsc --noEmit`
- `bun run format:check`


tests/bug-reports/bug-report-20260708T095725Z/__snapshots__/bug-report-20260708T095725Z.snap.svg
MustafaMulla29, line 1: Isn't this a regression? Can you explain me what is happening here

techmannih: @MustafaMulla29 in previous snapshot I see two issue, trace is interesting to the mpn and netlabel. In the new snapshot both are resolved

techmannih: Ohh yes it Miss the upper trace in the new snapshot. I am checking it why it is happening

MustafaMulla29 (approved):

Issue #674, opened 2026-07-17
JSON Bug Report:


mohan-bee:
### Solver input JSON attachment

[_board#665 _-SchematicTracePipelineSolver-solver-input.json](https://github.com/user-attachments/files/30108487/_board.665._-SchematicTracePipelineSolver-solver-input.json)

tscircuitbot:
Created or updated https://github.com/tscircuit/schematic-trace-solver/pull/675 for this JSON bug report.

Report id: `bug-report-20260717T022934Z`

DPS0340:
Diagnosed: 3 false junctions on this board come from `AvailableNetOrientationSolver` connector traces crossing other nets' traces (e.g. the V3_3 connector slicing through `JP1.1-SJ3.1` at (-7.1,-6.1)). Connector candidates are now checked against other-net traces in #703 — with a fallback so orientation constraints still take priority when every candidate is walled in. The 3 walled-in crossings on this board are pinned with a `test.failing` (input included) for a follow-up connector-rerouting pass.

PR #700, opened 2026-07-24
fix: merge generated net-label connector rails with same-net traces (#34)

DPS0340:
/claim #34

Fixes the remaining #34 symptom: close parallel same-net rails left unmerged when one of them is a **generated net-label connector trace**.

## Reproduce in an example file (as requested)

On current `main`, `example45` shows the exact symptom from the issue screenshot: the generated `SDA` net-label connector runs a vertical rail at `x=10.42` while the same-net `R10.2 -> CONN1.3` trace runs a parallel rail **0.166 apart** at `x=10.58`, with 0.5 of overlapping length — two rails that should be one line.

```
trace R10.2-CONN1.3               path: (10.15,0.70) -> (10.58,0.70) -> (10.58,0.10) -> (11.02,0.10)
trace available-net-orientation-1-SDA path: (10.37,0.70) -> (10.42,0.70) -> (10.42,0.20) -> (10.32,0.20)
```

## Root cause

The `aligning_same_net_rails` cleanup pass added in #663 (`traceCleanupSolver2`) restricts `eligibleTraceIds` to traces that existed at the **first** cleanup pass:

```ts
eligibleTraceIds: new Set(
  instance.traceCleanupSolver!.getOutput().traces.map((trace) => trace.mspPairId),
)
```

Net-label connector traces are generated *later* by `AvailableNetOrientationSolver`, so they're never eligible for rail alignment — the merging machinery from #663 simply never sees them.

## Fix

Add the generated `available-net-orientation-*` traces from the pre-alignment collision output to the eligible set. The existing rail-group merge (`alignSameNetRails`) then pulls them onto the shared coordinate — no changes to the alignment algorithm itself. All existing safety checks (obstacle collisions, label anchor preservation, readability scoring) still apply.

## Demonstrate fix before/after (as requested)

After the fix, the SDA connector rail merges onto `x=10.42` with the same-net trace:

```
trace R10.2-CONN1.3               path: (10.15,0.70) -> (10.42,0.70) -> (10.42,0.10) -> (11.02,0.10)
```

Snapshots updated for the three boards whose rails now merge: `example45`, `example49`, `bug-report-20260707T134549Z`. I scanned **all** example and bug-report boards with a close-parallel-rail detector before/after: example45's finding is eliminated and no new ones appear anywhere.

## Verification

- New regression test `tests/repros/repro-same-net-rail-merge-34.test.ts` asserts no close parallel same-net rails remain in example45 (fails on `main`)
- `bun test` — 141 pass, 0 fail
[...]

PR #703, opened 2026-07-24
fix: avoid net-label connector traces crossing other nets (#674)

DPS0340:
Progress on #674.

## Problem

`AvailableNetOrientationSolver` generates a connector trace from a label's original anchor to its corrected position, but never checks whether that connector **crosses an existing trace from a different net**. Every such crossing renders as a false junction on the schematic.

On the #674 board (`_board#665` input), 3 crossings appear:

```
JP1.1-SJ3.1 (net8)  x  available-net-orientation-3-V3_3   at (-7.1,-6.1)
U1.1-U1.3 (V3_3)    x  available-net-orientation-5-GND    at (-1.8,0.2)
available-net-orientation-4-V3_3  x  available-net-orientation-9-GND  at (-2.8,-5.5)
```

`example45` and `bug-report-20260708T055430Z` each had one more (visible in their snapshots as a GND connector slicing through a same-column V3_3 trace).

## Fix

`getCandidateStatus` now computes the candidate's connector trace and rejects the candidate (`trace-collision`) when the connector would cross another net's trace. A fallback pass in `processLabel` re-runs the search allowing crossings when no crossing-free candidate exists — so the orientation constraint still wins over crossing avoidance (this preserves repro51's pinned "orientation over purity" behavior; that test still passes unchanged).

## Results

- `example45`: crossings 2 → 1 (snapshot updated; the GND connector now attaches from the clear side)
- `bug-report-20260708T055430Z`: crossings 2 → 1 (snapshot updated)
- `#674` board: the shifted-anchor candidates for the remaining 3 are all walled in by another net's trace along the entire shift axis — fixing those requires routing the connector *around* the blocking trace, which is a bigger change. Pinned with a `test.failing` regression (board input included) so the remaining work is tracked by CI and this PR stays reviewable.

## Verification

- `bun test` — 141 pass, 0 fail (1 new intentional `test.failing`)
- `bunx tsc --noEmit` / `biome check` — clean

Happy to follow up with the connector-rerouting pass for the walled-in cases if this direction looks right.


DPS0340:
Following the repo's repro-first workflow: opened #706 as the standalone reproduction PR for #674 (pins the three false-junction crossings with a visual snapshot). This PR is the corresponding fix — happy to rebase once the repro lands.

DPS0340:
Rebased onto current `main` (through #714), no conflicts. Full suite passes (145 pass / 0 fail) and CI is green, so this is ready whenever you get to it.

DPS0340:
@mohan-bee thanks again for #700 — if you're up for one more, this is the next real fix in the same area (5 files).

I re-verified the numbers against current `main` just now, so they're not stale:

| board | main | this branch |
|---|---|---|
| `example45` | 2 crossings | **1** |
| `bug-report-20260708T055430Z` | 2 crossings | **1** |
| #674 board | 3 crossings | 3 (unchanged — see below) |

To be upfront about the limitation: the three crossings on the #674 board itself are **not** fixed here. Their shifted-anchor candidates are walled in by another net's trace along the whole shift axis, so resolving them needs the connector routed *around* the blocking trace rather than the anchor shifted. That's a bigger change, so I pinned them with a `test.failing` (board input included) to keep this PR reviewable and keep the remaining work visible in CI.

Full suite: 145 pass / 0 fail. Rebased onto current `main`, mergeable, CI green.

Happy to do the connector-rerouting pass as a follow-up if this direction looks right to you.

tests/bug-reports/bug-report-20260717-connector-crossing-674/input.json
mohan-bee, line 1: why no snapshots for this ?

mohan-bee (commented):
see comments

DPS0340:
@mohan-bee good catch — fixed in c18cc86.

The reason there was no snapshot: the existing assertion is `test.failing`, so the board never actually got rendered and no snapshot was written. I've added a passing render test alongside it, so the board is now pinned:

```
tests/repros/__snapshots__/repro-connector-trace-crossing-674.snap.svg
```

That makes the three remaining crossings visually reviewable instead of just being a number in an assertion, and it'll catch any visual regression while the connector-rerouting follow-up is still outstanding.

Full suite: 146 pass / 0 fail.

tests/bug-reports/bug-report-20260717-connector-crossing-674/input.json
mohan-bee, line 1: i still don't see anything .. for this input.json

DPS0340:
@mohan-bee you were right, and I misread your first comment — sorry for the round trip.

I thought you were asking me to *add* a snapshot, so I added one. The actual problem was that the `input.json` shouldn't have existed at all: the #674 issue already has a bot-imported bug report at `tests/bug-reports/bug-report-20260717T022934Z/` (the directory name comes from the issue's `created_at`, generated by `.github/scripts/import-json-bug-report.ts`). I'd hand-created a second directory containing a **byte-identical copy** of that same board — I diffed them to confirm.

Fixed in 5c8a733:

- deleted `tests/bug-reports/bug-report-20260717-connector-crossing-674/`
- repro now imports `bug-report-20260717T022934Z.json`
- dropped the snapshot I'd added, since that bug report already has its own

So no hand-made bug reports and no duplicated board. Full suite 145 pass / 0 fail.

---

Separately, e3cb03c actually fixes part of what this PR was leaving behind. The remaining crossings had a root cause I'd missed: labels are processed one at a time, so a connector is only validated against traces that exist at that moment. On this board `orientation-4` is routed at index 14 but collides with `orientation-9`, which isn't created until index 17 — the earlier connector never had a chance to see it.

Added a post-pass that runs once every connector exists and swaps in the mirrored L-route where that clears a crossing:

| board | before | after |
|---|---:|---:|
| #674 board | 3 | **2** |
| bug-report-20260706T213649Z | 11 | **9** |
| bug-report-20260706T220324Z | 6 | **5** |
| bug-report-20260721T221026Z | 5 | **4** |
| bug-report-20260717T022934Z | 3 | **2** |
| example43 | 1 | **0** |

The last 2 on the #674 board are still walled in — the mirrored route crosses something too, so those need the connector routed *around* the blocker. Still pinned by the existing `test.failing`.

Happy to split that post-pass into its own PR if you'd rather keep this one to the original scope.

DPS0340:
Cleaned this PR back to its original scope (a8b95d9). It's now just two commits:

1. the connector-crossing avoidance itself
2. the fix for your review — the hand-made bug report is gone, the repro imports `bug-report-20260717T022934Z` instead

The post-pass work I mentioned has been split out into **#716** as you'd expect, so this PR stays reviewable on its own. #716 is stacked on this branch, so this one goes first.

Worth mentioning what #716 turned up, since it changes the story here: I said the remaining crossings needed "routing around the blocking trace", and that turned out to be exactly right — for the last two, the blocker sits between the endpoints on *both* axes, so no L-shape avoids it. With detours added, the #674 board goes to **0 crossings** and the repro's assertion is no longer `test.failing`.

This PR on its own: full suite 145 pass / 0 fail, CI green.

DPS0340:
@MustafaMulla29 @ShiboSoftwareDev — adding you two; you're the top two mergers here (19 and 10 of the last ~60) and I'd only tagged @mohan-bee.

Small: 4 files, but only **2 are source** — `AvailableNetOrientationSolver.ts` (+58/-1) and the repro test (+62). The other two are snapshot refreshes.

Progress on #674. Net-label connector traces were being routed across other nets' traces. The connector is validated only against traces that exist at the moment it's routed, so a connector created early can't see one created later — concretely, `orientation-4` is routed at index 14 but collides with `orientation-9`, created at index 17.

This picks the mirrored L-route when the default one crosses. Measured different-net crossings vs `origin/main`:

```
example45                       2 → 1
bug-report-20260708T055430Z     2 → 1
```

The #674 board's own 3 crossings are unchanged here — those need the detour routing, which is split into #716 (stacked on this one) to keep this PR reviewable. #716 takes that board 3 → 0.

Full suite green, `MERGEABLE / CLEAN`.

DPS0340:
Closing this in favour of **#719**, which fixes the same root cause more effectively. I should have caught the duplication before opening both — apologies for the extra review surface.

Both PRs add the same missing check to `AvailableNetOrientationSolver`: a net-label's connector trace is validated against chips and other labels, but never against existing traces. They differ only in the fallback when no crossing-free candidate exists.

- **This PR (#703)** retries the entire search with the check disabled.
- **#719** rejects the candidate and lets the search continue to the next one.

I measured both against `origin/main` on all 19 imported bug reports, counting different-net crossings in the final trace geometry:

```
origin/main   TOTAL 40
#719          TOTAL 19
#703 (this)   TOTAL 39
```

Per board, this PR is worse on six of them:

| board | #719 | #703 |
|---|---:|---:|
| bug-report-20260706T213649Z | 4 | 11 |
| bug-report-20260706T220324Z | 4 | 6 |
| bug-report-20260707T134549Z | 2 | 3 |
| bug-report-20260716T144856Z | 0 | 4 |
| bug-report-20260717T022934Z | 0 | 3 |
| bug-report-20260721T221026Z | 2 | 5 |

The reason is that the fallback fires on nearly every label, so this PR degrades to the original behaviour and removes one crossing out of forty. The retry was meant to guarantee a label still gets placed, but in practice it just disables the fix.

They are also not stackable — merging both would leave the disabled-check path as dead code alongside a redundant scan.

So: **#719 is the one worth reviewing**, and it is best read after **#718**, which adds the counter both were measured with.

On @mohan-bee's review here — the snapshots and the bot-imported bug report were addressed in the commits above, and that work carries over to #719, so nothing is lost by closing this.

PR #704, opened 2026-07-24
test: repro #670 — labeled rail nets get long bus traces instead of net labels

DPS0340:
Reproduction PR for #670, following the repo's repro-first workflow (per the review guidance on #690).

Pins the current behavior on the #670 board (`_board#1332` input): `LongDistancePairSolver` connects pins on the **labeled** V3_3/GND rails with long orthogonal bus traces even though every one of those pins already receives a net label (V3_3 declared `y+`, GND `y-`):

```
JP1.2-JP2.2  len=12.7   (8.0,2.3)->(8.0,2.5)->(5.5,2.5)->(5.5,-7.6)
D2.2-JP6.1   len=11.3
U1.2-JP6.2   len=9.8
```

The test asserts exactly these three traces exist (with a visual snapshot), so a fix flips the assertion to `.toEqual([])`.

Proposed fix in #702 (separate PR): cap long-distance neighbor candidates at `maxMspPairDistance` for labeled nets with >2 pins — the same threshold `MspConnectionPairSolver` already uses when it decides a pair should stay label-connected.


DPS0340:
Superseded by #702, which contains this same regression test (`tests/repros/repro-long-trace-instead-of-netlabel-670.test.ts`) plus the actual fix. Closing this to keep the review queue focused on the PR that resolves #670.

PR #705, opened 2026-07-24
test: repro #655 — net labels are placed inside chip bodies

DPS0340:
Reproduction PR for #655, following the repo's repro-first workflow.

Pins the current behavior on `bug-report-20260707T230831Z` (already in the repo): the `VIN`, `PG`, `VOUT` and `U1.FB to R2.pin1` net labels end up **100% inside chip bodies**, rendering behind the components and illegible.

Root cause (as noted in the issue): `NetLabelNetLabelCollisionSolver.findNextCollidingPair` only looks for label-**label** overlaps, so a label inside a chip that isn't also colliding with another label is never relocated. A second layer: several chips on this board overlap each other, so a pin can be covered by *another* chip's body — port-only labels anchored on such pins collide in every orientation, meaning a fix also needs an anchor-escape strategy, not just a chip-collision pass.

The test asserts exactly the four buried labels, so a fix flips the assertion to `.toEqual([])`.

Proposed fix in #699 (separate PR): chip-collision relocation pass + outward anchor sliding for port-only labels. Verified there against all 21 bug-report boards (5 boards with buried labels on `main` → 0). Happy to rebase #699 on this repro once it lands.

### DIFF
--- a/tests/repros/repro-net-label-inside-chip-655.test.ts
+++ b/tests/repros/repro-net-label-inside-chip-655.test.ts
@@ -1,0 +1,74 @@
+import { expect, test } from "bun:test"
+import { SchematicTracePipelineSolver } from "lib/solvers/SchematicTracePipelineSolver/SchematicTracePipelineSolver"
+import "tests/fixtures/matcher"
+import inputProblem from "../bug-reports/bug-report-20260707T230831Z/bug-report-20260707T230831Z.json"
+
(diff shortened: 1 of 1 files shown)


DPS0340:
Superseded by #699, which contains this same regression test (`tests/repros/repro-net-label-inside-chip-655.test.ts`) plus the actual fix. Closing this so #699 is the single PR to review for #655.

PR #706, opened 2026-07-24
test: repro #674 — net-label connector traces cross other nets' traces

DPS0340:
Reproduction PR for #674, following the repo's repro-first workflow.

Pins the current behavior on the #674 board (`_board#665` input, attached to the issue): `AvailableNetOrientationSolver` places net-label connector traces without checking whether they cross traces from **other nets**. Each perpendicular crossing renders as a false junction. Three appear on this board:

```
JP1.1-SJ3.1 (net8)                x  available-net-orientation-3-V3_3   at (-7.1,-6.1)
U1.1-U1.3 (V3_3)                  x  available-net-orientation-5-GND    at (-1.8,0.2)
available-net-orientation-4-V3_3  x  available-net-orientation-9-GND    at (-2.8,-5.5)
```

The test asserts exactly these three crossings (with a visual snapshot), so a fix flips the assertion to `.toEqual([])`.

Proposed fix in #703 (separate PR): reject connector candidates that cross other nets during candidate evaluation, with an orientation-constraint-preserving fallback. That fix clears the same class of crossing on example45 and bug-report-20260708T055430Z; the three crossings here are the walled-in cases needing connector rerouting, documented there. Happy to rebase #703 on this repro once it lands.


DPS0340:
Superseded by #703, which contains this same regression test (`tests/repros/repro-connector-trace-crossing-674.test.ts`) plus the actual fix. Closing to avoid reviewing the same repro twice.

PR #708, opened 2026-07-24
test: repro #34 — generated connector rail runs parallel to same-net trace rail

DPS0340:
Reproduction PR for the remaining #34 symptom, following the repo's repro-first workflow.

The rail-alignment cleanup from #663 handles most same-net close-rail cases, but **generated net-label connector traces are excluded from the eligible set**: `traceCleanupSolver2` derives `eligibleTraceIds` from the *first* cleanup pass, which runs before `AvailableNetOrientationSolver` creates the connectors.

On `example45` (already in the repo) the generated `SDA` connector runs a vertical rail at `x=10.42` while the same-net `R10.2->CONN1.3` trace runs a parallel rail **0.166 away** at `x=10.58`, with 0.5 of overlapping length — two rails that should be one line:

```
R10.2-CONN1.3:                   (10.15,0.70)->(10.58,0.70)->(10.58,0.10)->(11.02,0.10)
available-net-orientation-1-SDA: (10.37,0.70)->(10.42,0.70)->(10.42,0.20)->(10.32,0.20)
```

The test asserts exactly this unmerged rail pair (with a visual snapshot), so a fix flips the assertion to `.toEqual([])`.

Proposed fix in #700 (separate PR, `/claim #34` there): add the generated `available-net-orientation-*` traces to the alignment pass's eligible set — the existing #663 merging machinery then pulls them onto the shared coordinate. Happy to rebase #700 on this repro once it lands.


DPS0340:
Superseded by #700, which covers this case with `tests/repros/repro-same-net-rail-merge-34.test.ts` plus the fix for #34. Closing this standalone repro.

PR #711, opened 2026-07-24
Add shared endpoint junction solver for schematic traces

mohan-bee:
Before: 

image

After: 

image


tests/examples/__snapshots__/example32.snap.svg
mohan-bee, line 1: there is a trace crossing ,,  i think we can do something , introducing something that checks the routes in the nearest possible way if that crosses with other net trace we can choose the other side e.g here top

tests/examples/__snapshots__/example40.snap.svg
mohan-bee, line 1: maybe same here too ,

tests/repros/__snapshots__/repro-netlabel-overlap-trace.snap.svg
mohan-bee, line 1: same here we can choose the top path too

lib/solvers/TraceCleanupSolver/sameNetRailAlignment/getRailGroups.ts
```diff
@@ -35,15 +35,20 @@ const corridorIsClear = (
   )
 }
 
+const getPinKey = (pin: SolvedTracePath["pins"][number]) =>
+  `${pin.chipId}::${pin.pinId}`
```
seveibar, line 39: 🤔

seveibar: you shouldn't do this-

seveibar (changes requested):
not a clear win, find a better method.

Don't introduce new types of ids

PR #718, opened 2026-07-25
test: pin different-net trace crossings per bug report

DPS0340:
## Why

Nearly every bug report in `tests/bug-reports/` is verified by an SVG snapshot. That catches *change*, but not *quality* — a trace crossing that appears or disappears is a handful of pixels in a large regenerated SVG, and easy to approve without noticing.

Different-net crossings are the thing several of these reports are actually about, so I measured them across every imported board:

| crossings | segments | board |
|---:|---:|---|
| 11 | 149 | bug-report-20260706T213649Z |
| 6 | 199 | bug-report-20260706T220324Z |
| 5 | 168 | bug-report-20260721T221026Z |
| 4 | 52 | bug-report-20260716T144856Z |
| 3 | 38 | bug-report-20260717T022934Z |
| 3 | 67 | bug-report-20260707T134549Z |
| 2 | 20 | bug-report-20260708T055430Z |
| 2 | 154 | bug-report-20260707T092615Z |
| 2 | 28 | bug-report-20260707T230831Z |
| 1 | 80 | bug-report-20260707T140410Z |
| 1 | 22 | bug-report-20260707T020342Z |
| 0 | — | (8 further boards) |

**40 crossings across 19 boards.** None of that was visible from CI before.

## What this adds

- `tests/fixtures/traceCrossings.ts` — `getTraceCrossings(traces)`, mirroring the shape of the existing `getTraceLabelCollisions` helper next to it. Counts only *proper* crossings between segments on different nets; shared endpoints on the same net are excluded, since traces legitimately meet there.
- `tests/different-net-trace-crossings.test.ts` — pins the count per board, one named test each.

**These numbers are the current state, not a target.** Several of these boards are open bugs and the counts should come down. The value is that they can't quietly go *up*: a routing or cleanup change that makes a board worse now fails an assertion that names the board, instead of sliding into a regenerated snapshot.

There's also a guard test asserting the pinned set matches what's on disk, so a newly imported bug report can't go unmeasured.

## Verification

- **The guard actually bites.** Changed the expected count for `bug-report-20260706T213649Z` from 11 to 10 and got `Expected: 10, Received: 11`. A pinning test that can't fail is worse than no test.
- Full suite: **164 pass / 0 fail / 4 skip** (was 144 before, +20 from this file).
- `tsc --noEmit` clean, biome clean.

One implementation note: I initially read traces from `netLabelNetLabelCollisionSolver`, the last pipeline step — but its `getOutput()` returns only `netLabelPlacements`. `tsc` caught it. Final trace geometry comes from `traceCleanupSolver2`, which is what this reads now.

## Follow-up

If this is useful, the natural next step is fixing the worst boards and lowering the numbers in the same PR. `bug-report-20260706T213649Z` at 11 is the obvious first target — happy to take that on.

**Diff: 2 new files, no existing file touched.**

### DIFF
--- a/tests/different-net-trace-crossings.test.ts
+++ b/tests/different-net-trace-crossings.test.ts
@@ -1,0 +1,74 @@
+import { expect, test } from "bun:test"
+import fs from "node:fs"
+import path from "node:path"
+import { SchematicTracePipelineSolver } from "lib/solvers/SchematicTracePipelineSolver/SchematicTracePipelineSolver"
+import { getTraceCrossings } from "tests/fixtures/traceCrossings"
+
+/**
+ * Different-net trace crossings, measured across every imported bug report.
+ *
+ * These are the current values, not a target — several of these boards are
+ * open bugs. The point is that the numbers are pinned: a routing or cleanup
+ * change that makes any board worse fails here instead of sliding into a
+ * regenerated SVG snapshot where a crossing is easy to miss.
+ *
+ * If a change legitimately improves a board, lower the number in the same PR.
+ */
+const EXPECTED_CROSSINGS: Record<string, number> = {
+  "bug-report-20260706T213649Z": 11,
+  "bug-report-20260706T220324Z": 6,
+  "bug-report-20260721T221026Z": 5,
+  "bug-report-20260716T144856Z": 4,
+  "bug-report-20260717T022934Z": 3,
+  "bug-report-20260707T134549Z": 3,
+  "bug-report-20260708T055430Z": 2,
+  "bug-report-20260707T092615Z": 2,
+  "bug-report-20260707T230831Z": 2,
+  "bug-report-20260707T140410Z": 1,
+  "bug-report-20260707T020342Z": 1,
+  "bug-report-20260717T031704Z": 0,
+  "bug-report-20260708T053736Z": 0,
+  "bug-report-20260707T134722Z": 0,
+  "bug-report-20260707T141025Z": 0,
+  "bug-report-20260708T095725Z": 0,
+  "bug-report-20260707T141421Z": 0,
+  "bug-report-20260717T042845Z": 0,
+  "bug-report-20260724T175257Z": 0,
+}
+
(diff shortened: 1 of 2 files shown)


DPS0340:
Updated this to pin **overlaps as well as crossings**, because I found the one-sided version scoring a good change as a bad one.

I had used the earlier version of this counter to report a "regression" from #727. It was not a regression. Counting different-net collinear overlaps too:

```
                        crossings   overlaps
ca7e80c (before #727)       40          1
be20aa7 (#727 merged)       41          0
```

Same pair of segments on both sides, on the single board that moved:

```
bug-report-20260707T092615Z
before:  overlap   C_SENSE.2-C1.2  ||  available-net-orientation-13-BAT_POS  (vertical, span 0.141)
after:   crossing  C_SENSE.2-C1.2  ×   available-net-orientation-13-BAT_POS  (x=-9.809, y=-1.764)
```

#727 separated two nets that had been drawn on top of each other and left them crossing at a point instead — the right trade, since an overlap looks like one trace and hides that two nets are there.

**That is the argument for pinning both.** A crossings-only guard is gameable in one specific direction: any crossing can be eliminated by making the two nets run along each other, and a one-sided counter reads that as an improvement. Pinning both makes the trade explicit — a PR that moves a board from `2/1` to `3/0` has to say so, rather than passing or failing depending on which quantity happened to be measured.

Still measurement-only — no `lib/` changes:

- `getTraceOverlaps` added next to `getTraceCrossings` in `tests/fixtures/traceCrossings.ts`. Shared endpoints are excluded, since traces meeting at a point is normal; only a shared *span* counts.
- the per-board table now pins `{ crossings, overlaps }`.

Verification, since a guard that cannot fail is worth nothing:

- checked out `ca7e80c` with the new fixture → `092615Z` fails on `overlaps: 1`, the other 19 pass. The counter detects the real state the crossings-only version missed.
- current `main` → all 20 boards pass.
- full suite `167 pass, 4 skip, 0 fail`; `bun run format:check` clean; `bunx tsc --noEmit` exits 0.

PR #724, opened 2026-07-25
Add MPL3115A2 VCC label placement repro

seveibar:
## Summary

- add the extracted schematic trace solver input for the SparkFun MPL3115A2 altitude/pressure sensor breakout
- pin the current behavior where the VCC label for U1.1 (VDD) and U1.4 (VDDIO) is constrained to `y+` but anchors left of the pin column (`x = -1.2` instead of `x = -1`)
- add a focused SVG solver snapshot with ratsnest guides hidden

## Why

This reproduces the case reported from [tscircuit/footprinter#728](https://github.com/tscircuit/footprinter/pull/728/files), corresponding to `boards/SparkFun-Altitude-Pressure-Sensor-Breakout-MPL3115A2/__snapshots__/SparkFun-Altitude-Pressure-Sensor-Breakout-MPL3115A2.circuit-schematic.snap.svg` in [tscircuit/sparkfun-boards](https://github.com/tscircuit/sparkfun-boards). The left-hand anchor makes the VCC connection appear to shoot out from VDDIO before turning upward. The semantic assertion and snapshot give a future placement fix a stable regression target.

## Validation

- `bun test` — 145 passed, 4 skipped
- `bunx tsc --noEmit`
- `bun run format:check`
- `git diff --check`


seveibar:
Closing because this did not reproduce the reported failure. The current SparkFun snapshot only detours 0.2 units before a y+ VCC label. The older far-left path in sparkfun-boards@2cf8838 came from an explicitly positioned netlabel (schX=-1.5, schY=0.6), not automatic solver placement. footprinter#728 also contains only FPC footprint changes, so there is no verified solver input associated with the original link.

PR #727, opened 2026-07-25
Avoid overlaps from generated net-label traces

imrishabh18:
## ISSUE

image

## Summary

- rebase onto the merged RP2040 USB-C repro from #728
- rerun overlap resolution after `AvailableNetOrientationSolver` creates connector traces
- constrain that late pass so only generated `available-net-orientation-*` traces may move
- update the merged repro assertion from the two known GND overlaps to no cross-net collinear overlaps

## Root cause

The pipeline resolved cross-net trace overlaps before net-label orientation correction. `AvailableNetOrientationSolver` then added the CC2 connector later, so that generated connector never went through overlap resolution and landed directly on two GND traces.

The long U-shaped trace highlighted in the repro is legitimate: it is `schematic_port_94-schematic_port_92`, the GND connection from R9’s downward-facing bottom pin to R8’s downward-facing bottom pin. A straight segment would cross both resistor bodies, so the line solver routes below R9, along the component-clearance boundary, and back into R8 from below.

The revised fix preserves that established GND route and shifts only the generated CC2 connector with endpoint-preserving terminal jogs.

## Tests

No additional repro test is added by this branch; it modifies the merged #728 test and snapshot. Two existing snapshots containing the same generated-label overlap behavior are refreshed.

## Related

Complements tscircuit/core#2878, which prevents different-net trace contacts from being rendered as electrical junctions.

## Validation

- `bun test` — 147 passed, 4 skipped
- `bunx tsc --noEmit`
- `bun run format:check`


seveibar (approved):

DPS0340:
@imrishabh18 — #727 looks like it improved one board and regressed another. Reporting with numbers since the regression isn't visible from the snapshots it updated.

Counting different-net trace crossings across all imported bug reports, immediately before and after #727:

```
ca7e80c (before #727)   TOTAL 40
be20aa7 (#727 merged)   TOTAL 41
```

The change is on one board:

```
bug-report-20260707T092615Z   2 → 3
```

Every other board is unchanged, and `repro-rp2040-usb-cc2-ground-overlap` — the case #727 targets — is genuinely fixed. So this is a narrow side effect of `TraceOverlapShiftSolver`, not a broad one.

I found it because **#718** (open, measurement-only) pins these counts per board, and it now fails on exactly this one. That PR is just a counter plus a table of current values — no behaviour change — and its whole purpose is to make this class of regression fail a named assertion instead of disappearing into a regenerated SVG. It would have caught this at review time.

Reproduce:

```bash
git checkout be20aa7
git checkout <718-branch> -- tests/fixtures/traceCrossings.ts
# then count getTraceCrossings() per board in tests/bug-reports/
```

Not asking you to revert — one crossing traded for a fixed overlap may well be the right call, and you have context I don't. But if the trade was unintentional, `092615Z` is where to look.

DPS0340:
Correcting myself: I called this a regression above, and that was wrong. It is a trade, and it goes in the right direction.

I had only been counting **crossings**. When I also counted different-net **collinear overlaps** — two nets drawn on top of each other along a shared span — the picture inverted:

```
                        crossings   overlaps
ca7e80c (before #727)       40          1
be20aa7 (#727 merged)       41          0
```

And on the one board that moved, it is the **same pair of segments** on both sides of the ledger:

```
bug-report-20260707T092615Z

before:  overlap   C_SENSE.2-C1.2  ||  available-net-orientation-13-BAT_POS   (vertical, span 0.141)
after:   crossing  C_SENSE.2-C1.2  ×   available-net-orientation-13-BAT_POS   (at x=-9.809, y=-1.764)
```

So #727 took two different nets that were being drawn as one line over a 0.141 span and separated them, leaving them crossing at a point instead. A crossing is visible and unambiguous; a collinear overlap looks like a single trace and hides that two nets are there. That is worth one crossing.

**The mistake is instructive, so I fixed it in the measurement rather than just retracting the comment.** A crossings-only guard is gameable in exactly this direction: any crossing can be "removed" by making the two nets run along each other instead, and a one-sided counter scores that as an improvement. #718 now pins **both** quantities per board, so a change that trades one for the other shows up as a deliberate, visible trade instead of passing or failing by accident.

Verified the new overlap counter actually detects the thing it claims to: checked out `ca7e80c` with the updated fixture and the `092615Z` case fails with `overlaps: 1`, while on current `main` all 20 boards pass. Full suite is green (`167 pass, 0 fail`), as are `format:check` and `tsc --noEmit`.

Sorry for the noise on a merged PR — the earlier comment made a claim my measurement did not support, and I would rather correct it in place than leave it standing.

PR #744, opened 2026-07-30
Fix schematic trace routing to form shared same-net load junctions

mohan-bee:
motivation: same net traces were being stepped which doesn't improving readability. 

The solver recognizes nearby loads connected to the same net and builds a shared rail:

Before: 

image

After:

image


MustafaMulla29:
@mohan-bee bun link this in core and update all the schematic snapshots, and see that all the snapshots are improving or regressing

mohan-bee:
> @mohan-bee bun link this in core and update all the schematic snapshots, and see that all the snapshots are improving or regressing

seen that is improving not any regression

seveibar (changes requested):
don't run a solver within an onSolved, use the pipeline properly

seveibar (approved):

PR #750, opened 2026-07-31
fix: prevent standalone net labels from intersecting traces

techmannih:


tests/bug-reports/bug-report-20260731T052229Z/__snapshots__/bug-report-20260731T052229Z.snap.svg
MustafaMulla29, line 1: Doesnt the trace to that label look wierd? 

Screenshot 2026-07-31 at 3 32 55 PM

cant we just push the below label, then it will have a straight trace.

MustafaMulla29 (approved):

mohan-bee (commented):
@MustafaMulla29  @techmannih that netlabel is bad ... They were not aligned properly still ... !!

MustafaMulla29:
> @MustafaMulla29 @techmannih that netlabel is bad ... They were not aligned properly still ... !!

This is a follow up improvement, the current output is acceptable

PR #754, opened 2026-08-01
Codespace sturdy guacamole r4j7qpgqp74qhxjgr

khozakhulile27-netizen:


khozakhulile27-netizen (commented):
feat: simplify collinear trace vertices + prep same-net rail merging

PR #755, opened 2026-08-02
Align blocked vertical net labels

mohan-bee:
### Motivation
- This is a follow up improvement of #750 
- Keep corrected vertical net labels visually aligned and separated.

### Before
- VDDD was slightly outside the VDDR and VDDS column.
- VDDIO touched the corrected VDDD label.

### After
- VDDD aligns with VDDR and VDDS.
- VDDIO keeps a small gap from VDDD.


tests/bug-reports/bug-report-20260731T052229Z/__snapshots__/bug-report-20260731T052229Z.snap.svg
techmannih, line 1: @mohan-bee I think both are different case, top two netlabel are muti pin netlabel and  last netlabel is single pin netlabel, so the distance should be same from the pin point or the junction points.  Wdyt?

mohan-bee: no, every netlabel probably will have only one pin .. the connection differs . somehow we need to align them in a proper vertically .. to keep our schematic look cleaner .. tbh .. not a big change ..

lib/solvers/AvailableNetOrientationSolver/AvailableNetOrientationSolver.ts
```diff
@@ -51,6 +51,10 @@ export class AvailableNetOrientationSolver extends BaseSolver {
   currentLabel: NetLabelPlacement | null = null
   currentCandidateResults: EvaluatedCandidate[] = []
 
+  private blockedStandaloneLabelAlignment: {
+    labelIndex: number
+    columnX: number | null
```
MustafaMulla29: What is columnX? Can you type the `blockedStandaloneLabelAlignment` better or maybe you split that obj into diff variables.(Whichever is easier which follows codebase patters)

mohan-bee: yea, very bad , will see a different option

mohan-bee:
@MustafaMulla29  @techmannih  take a look

MustafaMulla29 (approved):

PR #772, opened 2026-08-04
Avoid shared-pin trace spikes

mohan-bee:
### Motivation
- Shared component pins could make a solved trace borrow a sibling connection's label width and route around inflated text padding.

Before: 

image

After : 

image


seveibar (approved):
not quite right 
image

mohan-bee:
> not quite right image

will make a followup

PR #785, opened 2026-08-06
Prefer shorter valid schematic label detours

MustafaMulla29:
Screenshot 2026-08-06 at 5 26 54 PM

## Summary

- generate a compound candidate for simple five-segment elbows instead of stacking local four-point notches
- rank valid reroutes by Manhattan length, then rendered-label overlap count, then point count
- reject simple candidates that move an existing net-label anchor or transfer the trace into more rendered labels
- add an exact regression test for the U1.pin1 route from core repro148

## Root cause

The label rerouter only generated local detours around the label currently being processed. In the DRV8323 case, repeated passes stacked those detours into the long U-shaped VM route.

The first fix generated the missing simple route, but candidate selection still prioritized removing a generated label overlap over path length. In core repro148 that made U1.pin1 take a longer four-point notch around a GND label that core does not ultimately render.

## Fix

`generateSimpleElbowDetourCandidates` produces the missing compound route by moving the elbow transition and then shifting the colliding segment beyond the neighboring label.

`SingleOverlapSolver` admits that route only when it:

1. reduces overlaps across the complete rendered-label set;
2. preserves existing label anchors;
3. has the same Manhattan length as a standard detour candidate;
4. uses fewer points than that standard candidate.

All candidates then use the same deterministic ordering: shortest path first, fewer label overlaps second, and fewer points third. The existing target-label, obstacle, other-trace, retry, and solver-lifecycle checks remain in place. There are no circuit IDs or caller-controlled strategy flags.

## Result

| Case | Before | After | Collision result |
| --- | ---: | ---: | --- |
| DRV8323 VM rail | 14 points / 14.322 length | 6 points / 12.002 length | no overlap |
| DRV8323 neighboring GND | 8 points / 10.582 length | 6 points / 7.382 length | same existing C_CP/CPL overlap |
| Full DRV8323 output | 38 points / 50.224 length | 28 points / 44.704 length | same single overlap |
| Core repro148 U1.pin1 | 10-point label notch | original 6-point route | no added rendered collision |

## Snapshot audit

Every changed existing snapshot was compared by path length, point count, collision identity, crossings, and visual geometry:

- `bug-report-20260806T093501Z`: 5.52 shorter and 10 fewer points; final overlap set is unchanged
- `bug-report-20260706T220324Z`: two routes are 1.362 and 1.15 shorter; no added overlap
- `example33`: same length with 8 to 6 points and 6 to 4 bends
- `bug-report-20260706T213649Z`: one route drops from 8 to 6 points at the same length; an equivalent six-point route now remains at its upstream position instead of being replaced by an equal-score candidate; length, bends, crossings, and label-overlap count do not worsen
- `bug-report-20260707T092615Z`: unchanged from `main`
- new `repro-core-148-pin1-label-detour` fixture: asserts the exact clean six-point U1.pin1 path

## Validation

- linked the local solver build into current core and regenerated repro148; the U1.pin1 notch is gone
- `bun test` — 170 passed, 4 skipped, 0 failed
- `bunx tsc --noEmit`
- `bun run build`
- `bun run format:check`
- `git diff --check`


lib/solvers/TraceLabelOverlapAvoidanceSolver/rerouteCollidingTrace.ts
imrishabh18, line 1: Not the right place to add this method, this is not linked to the collidingTrace.

imrishabh18 (approved):
Move this to separate post processing step which simplifies the trace which are having many elbow transitions

PR #786, opened 2026-08-06
Keep traces clear of net label side walls

mohan-bee:
### Motivation
- Traces can run on or near the long side walls of their own net labels, obscuring the label boundary.

image


seveibar (changes requested):
net label should be on top of trace

PR #791, opened 2026-08-08
Replace crossing GND traces with net labels

mohan-bee:
### Motivation
- GND routes can cross signal traces even when the connection can be represented safely with net labels.

### Before

image

### After

image


lib/solvers/AvailableNetOrientationSolver/AvailableNetOrientationSolver.ts
```diff
@@ -43,11 +44,13 @@ import type {
 import { visualizeAvailableNetOrientationSolver } from "./visualize"
 
 const LABEL_TRACE_CLEARANCE = 0.1
+const GROUND_NET_ID = "GND"
```
seveibar, line 47: NO STRING MATCHING!!!!!!!!!!!!!!

PR #793, opened 2026-08-08
Slide overhanging inline net labels instead of falling back

seveibar:
Follow-up to #792, addressing [review feedback on core#3085](https://github.com/tscircuit/core/pull/3085): in the U1→LED1 scenario the inline label fell back to an anchored tag — which renders vertically across the very wire the inline label was meant to keep clean.

## Why it fell back

`USER_LED_ANODE` needs 1.8 units; every straight run of the routed trace is ~0.95. Runs it can't fully fit on were only ever tried at their **midpoint**, and from the midpoint the label always collided with U1's or LED1's chip box — so placement declined entirely.

## Fix

Candidates now slide in the doesn't-fit case too. The slide range is `|segmentLength − labelWidth| / 2` about the midpoint in **both** cases:

- label shorter than the run → slides while staying fully on the wire (unchanged behavior);
- label longer than the run → slides while keeping the wire fully under the label, shifting the overhang toward whichever end is open.

Obstruction checks are untouched — the label still never sits on a chip, component text, or a foreign-net trace.

## Result

| before (#792) | after |
|---|---|
| anchored tag crossing the wire | label runs along the wire between the chips |

label overhangs the short run, clearing both chips

New test uses the exact input problem dumped from the core scenario (soic8 + 0603 LED, 3 units apart) and asserts the label is placed and overlaps neither chip box.

Candidate generation is byte-identical for labels that fit (`halfRange` equals the old `slack/2`), so no existing snapshots changed. `bun test`: 175 pass / 0 fail. Typecheck clean.

🤖 Generated with [Claude Code](https://claude.com/claude-code)

### DIFF
--- a/lib/solvers/InlineNetLabelSolver/InlineNetLabelSolver.ts
+++ b/lib/solvers/InlineNetLabelSolver/InlineNetLabelSolver.ts
@@ -374,9 +374,11 @@ export const estimateInlineNetLabelWidth = (
  * middle of the segment outwards. Sliding the label along the wire lets it
  * dodge a chip that only crowds one end.
  *
- * When the label is shorter than the segment, candidates are limited to
- * positions that keep it fully on the wire; when it is longer, only the
- * midpoint is offered (it will overhang either way).
+ * When the label is shorter than the segment, candidates keep it fully on the
+ * wire; when it is longer, candidates keep the wire fully under the label (the
+ * overhang shifts between the two ends). Either way the slide range is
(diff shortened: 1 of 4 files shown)


seveibar:
Closing in favor of a span-based approach: overhanging a single run reads as floating text rather than a name sitting on its wire. Instead, a route that is straight except for small jogs (like the 0.3-unit elbow here) will be treated as one span, with the label centered over the route's overall extent and never extending past its endpoints. PR to follow.

PR #825, opened 2026-08-19
Fix same-net junction label alignment

MustafaMulla29:
## Summary

- restore valid same-net junction alignment for the board-589 regulator repro
- move only net labels attached to a rerouted trace using the existing label-movement logic
- keep moved label anchors on the aligned trace and validate candidates against the moved placements
- add regression coverage for label attachment, the shared horizontal rail, and orthogonal trace segments

## Root cause

#806 prevented the label from becoming detached by rejecting the alignment entirely. That left a visible step in the same-net rail. The alignment should remain valid; its attached label needs to move with the rerouted trace.

## Impact

The affected V3_3 branches now share one straight horizontal rail. The label remains attached, and the fix does not introduce diagonal trace segments.

## Validation

- `bun run build`
- `bun test` — 184 passed, 4 skipped, 0 failed
- `bun test tests/repros/repro-board-589-regulator-section.test.ts`
- linked `@tscircuit/core` TypeScript check
- affected core schematic snapshot rendered and reviewed


tests/bug-reports/bug-report-20260815T073240Z/__snapshots__/bug-report-20260815T073240Z.snap.svg
mohan-bee, line 1: Image

this looks bit bad

mohan-bee (approved):

tests/bug-reports/bug-report-20260815T073240Z/__snapshots__/bug-report-20260815T073240Z.snap.svg
MustafaMulla29: Will take a look

PR #838, opened 2026-08-20
reproduce board 1273 trace overlap cycle

mohan-bee:
motivation: this input is taking forever or getting timeout for routing


tests/repros/board-1273-trace-overlap-cycle.test.ts
```diff
@@ -0,0 +1,26 @@
...
+    await Bun.file(
+      new URL(
+        "./assets/board-1273-trace-overlap-cycle.input.json",
+        import.meta.url,
+      ),
+    ).text(),
+  ) as InputProblem
+  const solver = new SchematicTracePipelineSolver(inputProblem)
+
+  solver.solveUntilPhase("traceOverlapShiftSolver")
```
mohan-bee, line 19: this is where upto the board gets to route , after this gets stuck into a loop

PR #860, opened 2026-08-24
Separate inline fallback label width from routing

MustafaMulla29:
Screenshot 2026-08-24 at 8 36 04 PM

## Summary

- add fallbackNetLabelWidth for the conventional anchored label used only when inline placement fails
- centralize net-label width lookup across placement and orientation stages
- explicitly exclude fallback-only width from trace-line routing decisions
- add a focused TB67S579FTG-derived regression proving five routable signals remain continuous

## Context

This is the prerequisite for tscircuit/core#3405. Core needs to provide the full rendered width of an AGC_OUT anchored fallback label so an unrelated GND trace can avoid it. Reusing netLabelWidth for that purpose also enlarged routing-time text padding and caused several right-side traces to fail and become paired terminal stubs.

The separate fallback field preserves the existing routed traces while giving anchored fallback placement and collision stages the correct geometry.

## Testing

- bun test: 205 pass, 4 existing skips, 0 fail
- bunx tsc --noEmit
- bun run build
- focused solver snapshot and semantic assertions for routed and fallback cases

### DIFF
--- a/lib/solvers/AvailableNetOrientationSolver/AvailableNetOrientationSolver.ts
+++ b/lib/solvers/AvailableNetOrientationSolver/AvailableNetOrientationSolver.ts
@@ -18,7 +18,6 @@ import type {
 import { dir, type FacingDirection } from "lib/utils/dir"
 import { getNetLabelWidthForConnection } from "lib/utils/getNetLabelWidthForConnection"
 import { rectIntersectsAnyTextBox } from "lib/utils/textBoxBounds"
-import { AvailableNetOrientationObstacleIndex } from "./AvailableNetOrientationObstacleIndex"
 import {
   EPS,
   LABEL_SEARCH_STEP,
@@ -48,6 +47,7 @@ import type {
   EvaluatedCandidate,
 } from "./types"
 import { visualizeAvailableNetOrientationSolver } from "./visualize"
+import { AvailableNetOrientationObstacleIndex } from "./AvailableNetOrientationObstacleIndex"
 
 const LABEL_TRACE_CLEARANCE = 0.1
 
--- a/lib/solvers/AvailableNetOrientationSolver/AvailableNetOrientationSolver.ts
+++ b/lib/solvers/AvailableNetOrientationSolver/AvailableNetOrientationSolver.ts
@@ -16,7 +16,9 @@ import type {
   InputProblem,
 } from "lib/types/InputProblem"
 import { dir, type FacingDirection } from "lib/utils/dir"
+import { getNetLabelWidthForConnection } from "lib/utils/getNetLabelWidthForConnection"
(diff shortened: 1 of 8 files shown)


lib/solvers/AvailableNetOrientationSolver/AvailableNetOrientationSolver.ts
```diff
@@ -1292,26 +1293,11 @@ export class AvailableNetOrientationSolver extends BaseSolver {
   }
 
   private getNetLabelWidth(label: NetLabelPlacement) {
```
imrishabh18: redundant method

lib/solvers/SchematicTraceLinesSolver/SchematicTraceSingleLineSolver2/SchematicTraceSingleLineSolver2.ts
```diff
@@ -240,25 +241,13 @@ export class SchematicTraceSingleLineSolver2 extends BaseSolver {
   }
 
   private getNetLabelWidthForConnectionPair(netId: string) {
```
imrishabh18: redundant method

lib/utils/getNetLabelWidthForConnection.ts
```diff
@@ -0,0 +1,62 @@
...
+}
+
+export const getNetLabelWidthForConnection = ({
+  inputProblem,
+  netId,
+  pinIds,
+  includeFallbackNetLabelWidth = true,
+}: {
+  inputProblem: InputProblem
+  netId?: string
```
imrishabh18: Don't use `string` this creates confusion

imrishabh18 (changes requested):

lib/solvers/AvailableNetOrientationSolver/AvailableNetOrientationSolver.ts
MustafaMulla29: Removed the wrapper and now call `getNetLabelWidthForConnection` directly at its use sites.

lib/solvers/SchematicTraceLinesSolver/SchematicTraceSingleLineSolver2/SchematicTraceSingleLineSolver2.ts
MustafaMulla29: Removed the wrapper and call `getNetLabelWidthForConnection` directly while computing routing-time text-box padding.

lib/utils/getNetLabelWidthForConnection.ts
MustafaMulla29: Updated the helper parameter to use the existing `NetId` type.

imrishabh18 (approved):

PR #861, opened 2026-08-24
Keep anchored label widths out of inline trace routing

MustafaMulla29:
Screenshot 2026-08-25 at 1 03 09 AM

## Summary
- apply automatic fallback label widths only to port-only label placements
- keep explicit net label widths unchanged for routed labels
- use the same port-only distinction when repositioning or checking anchored labels
- assert that fallback widths do not change routed trace paths while skipped routes retain measured fallback bounds

## Root cause
The measured fallback width was also sizing temporary anchored labels on already-routed inline connections. Those temporary boxes became cleanup obstacles and moved unrelated LO2 and SERIAL_CLK elbows in the core TB67S579FTG repro. Port-only AGC_OUT labels still need the measured width, so they continue to reserve their real bounds and GND correctly routes around them.

## Verification
- bun test (206 pass, 4 skip)
- bun run build
- bun run format:check
- core TB67S579FTG repro rendered locally: AGC_OUT remains fixed, GND avoids its label, and the right-side routes retain their previous geometry


lib/solvers/AvailableNetOrientationSolver/AvailableNetOrientationSolver.ts
```diff
@@ -185,6 +185,8 @@ export class AvailableNetOrientationSolver extends BaseSolver {
       inputProblem: this.inputProblem,
       netId: blockedLabel.netId,
       pinIds: blockedLabel.pinIds,
+      includeFallbackNetLabelWidth:
```
seveibar: the word fallback is usually banned fwiw, i didn't approve it's initial introduction, maybe it's ok here but most likely this is a huge slop mistake imo CC @imrishabh18 who i think approved. What the fuck is a fallback width

seveibar (approved):

lib/solvers/AvailableNetOrientationSolver/AvailableNetOrientationSolver.ts
MustafaMulla29: fallbackNetLabel is a label which the solver didn't solve and falls back to the old style of netlabel instead of trace inline labels. My bad maybe I should do better naming

MustafaMulla29: You are right—the boolean was leaking a representation choice into every width lookup. I replaced the field with `anchoredNetLabelWidth` and split the width lookups so routing never receives anchored-label geometry; placement code requests it only when it is sizing an anchored label. I also renamed the focused test/snapshot. Full tests, typecheck, formatting, and build pass in 30d934b.

MustafaMulla29: Follow-up: I simplified this further. The existing `getNetLabelWidthForConnection` is now the only width getter; its old boolean is replaced by required `labelType: "inline" | "anchored"`. There are no wrapper getters or class helper. The branch is current with `main`, and full tests, typecheck, formatting, and build pass at 9c39cb0.

PR #879, opened 2026-08-25
Align same-net rails without moving labels

MustafaMulla29:
## Summary

- Preserve fixed outward net-label anchors while aligning same-net rail chains through shared pins.
- Align complete capacitor chains to the uniquely labeled terminal rail, including all three branches in the July 30 regression.
- Generate the three-pin GND connector from the adjacent rail bend without changing longer multi-pin rail ordering.
- Add exact regressions for the top capacitor rail, unchanged C26 rail, three-pin GND connector, and remote Type-C GND label.

## Implementation

- Same-net alignment remains contained in `TraceCleanupSolver/sameNetRailAlignment`.
- Existing nearest-endpoint rail groups retain their original behavior and ordering.
- Tied endpoint associations are considered only for local connections and only when a fixed label anchors the expanded group.
- Fixed labels may override readability scoring only for single-rail traces. Per-connection locality is retained unless the label is carried by an intentional multi-turn routed backbone.
- Connector selection stays in `AvailableNetOrientationSolver`; the nearest bend tie-break applies only when the outward row belongs to a trace directly adjacent to the label host.
- Repeated rail states terminate deterministically, preventing endpoint-component oscillation and maximum-pass parity changes.
- No post-processing was added to `Example28Solver` or `labelMovement.ts`.
- No bug-specific coordinates, component IDs, net names, generated-trace prefix checks, or duplicated path-intersection geometry are used.

## Blast-radius audit

- Regenerated the full suite and inspected every remaining snapshot structurally and visually.
- Reduced the snapshot diff from 18 files to 5.
- Restored `repro-type-c-vbus-led-ground-label`, `repro-bq25895-cross-net-trace-overlap`, `bug-report-20260707T141421Z`, `bug-report-20260728T144234Z`, `repro-tida010076-page02`, `board-1273-trace-overlap-cycle`, and `bug-report-20260721T221026Z` exactly to `main`.
- The five remaining snapshots contain actual same-net rail joins; the target snapshot also removes the coincident portion of the three-pin GND connector.
- Net-label anchors and centers remain fixed.
- Alignment candidates are rejected if they add turns, other-net crossings, component collisions, label overlaps, extra-elbow rewrites, or non-local label pulls.

## Verification

- `bun test --timeout 20000` — 213 passed, 4 skipped, 0 failed
- `bunx tsc --noEmit`
- `bun run build`
- `bun run format:check`


MustafaMulla29:
@seveibar not sure if this is good, it works for some cases, but it doesnt work for other similar cases.

mohan-bee:
i think i have a related PR @MustafaMulla29  i was polishing some regressions https://github.com/tscircuit/schematic-trace-solver/pull/859 check is this solving the same issue

seveibar (approved):
GREAT improvement

seveibar:
@mohan-bee we DO NOT use first-come first-serve merging. that is a policy of SLOW COMPANIES

PR #880, opened 2026-08-25
Allow nested solvers to use their own iteration budgets

ShiboSoftwareDev:
## Summary

- track parent-solver work separately from iterations delegated to an active child solver
- let each child solver use its own independent iteration budget
- retain the parent iteration guard for work performed outside child solvers
- update the PMP11282 repro from 22 queued pairs and a failed pipeline to a completed queue and solved pipeline

## Stack

This is the second PR in the stack and is based on the repro-only PR #883.

- #883: captured PMP11282 input, pre-fix assertions, and SVG snapshot only
- this PR: solver fix, post-fix assertions/snapshot, and focused `BaseSolver` unit coverage

## Validation

- `bun test` — 214 passed, 4 skipped
- `bunx tsc --noEmit`
- `bun run format:check`
- PMP11282 pipeline completes all 24 stages with no connection pairs left queued

This PR intentionally addresses only nested iteration-budget accounting. Other trace geometry and net-label issues remain out of scope.


seveibar (changes requested):
i don't think this makes sense

seveibar:
i don't think this is a good change, and it takes us further from solver-utils. Increase the parent budget or decrease the child budget, or figure out why it takes so long etc.

seveibar (commented):
is this schematic based on the ti schematic? is the layout accurate

ShiboSoftwareDev:
Closing because this changes BaseSolver's global iteration semantics to accommodate an invalid PMP11282 fixture. The underlying schematic forced distant nets into traces by using schMaxTraceDistance=100mm with automatic net labels disabled; the correction belongs in the TI reference first.

PR #883, opened 2026-08-25
Add PMP11282 parent trace budget reproduction

ShiboSoftwareDev:
## Summary

- add the full 113-chip `InputProblem` captured from the PMP11282 isolated DC/DC schematic
- reproduce `SchematicTraceLinesSolver` exhausting its parent iteration budget while 22 connection pairs remain queued
- commit the pre-fix semantic/debug SVG snapshot using the repository matcher

## Scope

This is the first PR in a two-PR stack and contains only the reproduction, assertions for the current behavior, and its SVG snapshot. It intentionally contains no solver implementation changes.

The follow-up fix is stacked in #880.

## Validation

- `bun test tests/repros/repro-pmp11282-iteration-budget.test.ts`
- `bun test` — 212 passed, 4 skipped
- `bunx tsc --noEmit`
- `bun run format:check`


seveibar (approved):

ShiboSoftwareDev:
Closing because the reproduction encoded an invalid PMP11282 configuration (100mm trace distance with automatic net labels disabled), so it is not representative of the intended schematic behavior. We will recapture a focused reproduction only if a real defect remains after restoring normal net-label fallback.

PR #891, opened 2026-08-25
Add PMP11282 endpoint net-label reproduction

ShiboSoftwareDev:
## Summary

- add a full-pipeline PMP11282 reproduction after #890
- use the same source-derived isolated DC/DC schematic input as the existing stack
- capture the remaining net-label output after all route pairs are attempted
- assert that 106 labels are placed, including 80 endpoint-pair identifiers across 62 generated nets
- check in the repository-standard semantic SVG snapshot

This is a reproduction-only PR. It does not change production behavior.

Stacked on #890.

## Validation

- `bun test tests/repros/repro-pmp11282-endpoint-netlabels.test.ts`
- `bunx tsc --noEmit`
- `bun run format:check`

The targeted reproduction passes. The full-suite check currently inherits the two
known failures introduced earlier in the stack at #888:

- `routes the usb power vbus connection without a label detour`
- `board 1273 solves within the trace overlap iteration limit`


ShiboSoftwareDev:
Superseded by the single complete PMP11282 reproduction in #887, which now captures the trace and downstream endpoint-label output in one generated snapshot.

PR #900, opened 2026-08-26
Avoid opposite-endpoint component text in terminal routes

ShiboSoftwareDev:
## Summary

- keep the existing text-clearance exception for terminal trace segments
- retain the actual text rectangle as a hard obstacle when a terminal segment wraps across its opposite endpoint component
- route the reduced PMP11282 U503.pin2 → TP502.pin1 connection below `PC817X4NSZ0F` instead of through it
- flip the focused collision assertion and update its SVG snapshot
- update the full PMP11282 reproduction so the same route change is visible in the original fixture

This is stacked on the reproduction-only PR. The production change is confined
to terminal-segment text-obstacle filtering in `SchematicTraceSingleLineSolver2`.

## Other affected snapshots

Three existing broad fixtures contained the same opposite-endpoint text
crossing and therefore receive the same correction:

- board 1273: route crossed the opposite endpoint's `U7` reference text
- bug-report-20260730T061837Z: route crossed the opposite endpoint's `U7` reference text
- TIDA-010076 page 02: route crossed the opposite endpoint's `J2` reference text

## Validation

- `bun test` — 217 pass, 4 skip, 0 fail
- `bunx tsc --noEmit`
- `bun run format:check`


tests/repros/__snapshots__/repro-pmp11282-isolated-dcdc.snap.svg
imrishabh18, line 1: It's difficult to review what's happening in here

ShiboSoftwareDev: @imrishabh18 most circuits are like this in this repo, that's why I made a focused repro in tests/repros/__snapshots__/repro-pmp11282-endpoint-trace-through-component-text.snap.svg

imrishabh18 (approved):
update core after this, to know if this change is breaking some snapshots or not

PR #910, opened 2026-08-26
Place opted-in inline labels on multi-pin net connections

MustafaMulla29:
## Summary

- try both sides of routed traces and terminal stubs before falling back to anchored labels
- derive inline placements for opted-in multi-pin nets from solved trace-connected components
- atomically replace short routed components only when every pin in the component remains represented
- treat `available-net-orientation-*` traces as generated anchored-label connectors, not routed circuitry, so converted endpoints receive real inline terminal stubs
- preserve inline terminal rows by moving the row outward when a retained anchored label cannot safely move, with collision checks for labels, traces, components, and text
- roll back an inline conversion for the whole net only when neither representation can be moved safely
- exercise the merged DRV8305 motor-driver repro with rats-net rendering hidden

## Verification

- `bun run format`
- `bun run build`
- 20 focused InlineNetLabelSolver, motor-driver bug-report, and USB regression tests passed
- the RP2040 Board Headers input produces inline D10/SCLK/MISO/MOSI labels at both endpoints with zero matching anchored labels
- linked core snapshots confirm the capacitor, section-autolayout, RP2040, repro125, and repro169 duplicate-label cases no longer render both styles

## Blast radius

- The existing USB power repro snapshot is unchanged from `main`; its high-speed pair remains consistently anchored because a partial inline conversion would overlap the retained peer label.
- The only existing solver snapshot changed from `main` is the targeted DRV8305 motor-driver bug report.
- Production logic contains no circuit-specific net names, component names, pin IDs, or schematic-port IDs. Decisions use generated-trace provenance, connectivity, component-side grouping, and geometric collision checks.


seveibar (commented):
Based on the title this is the wrong repo, core explicitly specifies inline net labels for <2 pin nets

MustafaMulla29:
> Based on the title this is the wrong repo, core explicitly specifies inline net labels for <2 pin nets

I have the core pr as well

seveibar (approved):

PR #917, opened 2026-08-26
Anchor vertical ground labels to clear component-side rails

mohan-bee:
**motivation**
the gnd label could be placed vertically instead of branched away

**before**
the hdc2080 reproduction kept the ground label on an outward column instead of the connected rail, leaving the label anchor and rail at different x coordinates. this real before schematic svg preserves the offset label.

image

**after**
the orientation solver accepts the direct host-trace candidate only when its label bounds remain on one component side and clear every unrelated pin; the ground label and rail now both anchor at x = 0.1 with exactly 1 ground rail trace. the focused test, type check, build, and all 217 tests pass, and this real after schematic svg shows the aligned label.

image


MustafaMulla29 (commented):
There are other tests with the same GND label cases, can you figure why didnt they improve

seveibar (approved):

PR #918, opened 2026-08-26
Align capacitor rail chains without blocking junction cleanup

mohan-bee:
### Motivation
- Keep the TAS2505 capacitor chain on one rail without losing other routing improvements.

### Before
- Whole-trace protection blocked perpendicular junction cleanup in RS485 and TPS61222.

### After
- Exact rail constraints preserve improved snapshots while RS485 and TPS61222 retain their previous layouts and TPS61222 bottom rail Y.


seveibar (commented):
see regressions

PR #919, opened 2026-08-26
Prevent grouped ground rails from crossing component rows

mohan-bee:
**motivation**
the msp stage must use authoritative ground metadata when grouped component rows share one electrical net so it does not create a long cross-row trace through the component groups.

**before**
the bluetooth reproduction produced 6 ground pairs, including 1 cross-row pair that joined the two local ground rails through the grouped components. this real before schematic svg preserves the crossing.

image

**after**
the minimum spanning tree forbids cross-row edges only when typed ground metadata and both grouped rows are present; the result has 5 local pairs and 0 cross-row pairs, while the non-ground control retains 6 pairs and 1 cross-row pair. core #3448 supplies the metadata, the focused tests, type check, build, and all 219 tests pass, and this real after schematic svg shows the separated rows.

image


mohan-bee:
@copilot  rerun ci checks

seveibar (approved):

PR #924, opened 2026-08-26
Repro PMP11282 net-label connector crossing component text

ShiboSoftwareDev:
## Summary\n\n- reduce the PMP11282 Q504/H501 geometry to the generated GND1 connector failure\n- preserve the original pin, component-text, label dimensions, and required y- orientation\n- add a semantic solver SVG snapshot generated from the TypeScript reproduction\n- assert the current connector crosses H501 value text\n\n## Stack\n\nThis reproduction is based on the retained endpoint-text fix PR. It intentionally contains no production-code changes; the follow-up PR is stacked on this branch.\n\n## Validation\n\n- bun test\n- bunx tsc --noEmit\n- bun run format:check

PR #925, opened 2026-08-26
Route generated net-label connectors out of source text

ShiboSoftwareDev:
## Summary\n\n- detect when a generated net-label connector starts inside one or more component text boxes\n- leave the containing text bounds through the nearest edge with wick clearance before continuing the orthogonal route\n- keep normal connector routing unchanged when the source is already clear\n- update both the focused reproduction snapshot and the original PMP11282 snapshot\n\n## Why\n\nIn PMP11282, Q504.pin3 begins inside the H501 value-text bounds, so some initial overlap is unavoidable. The old orientation-biased connector traveled laterally across the value. The new route takes the shortest exit downward, then stays outside the text while reaching the GND1 label.\n\n## Stack\n\nThis fix is stacked on the reproduction-only PR #924.\n\n## Validation\n\n- 218 passed, 4 skipped, 0 failed\n- bunx tsc --noEmit\n- bun run format:check


ShiboSoftwareDev:
Closing this approach: because the connector source is already inside the text bounds, nearest-edge escape is not an unambiguous collision fix. I am replacing it with a different focused PMP11282 issue.

PR #928, opened 2026-08-26
Add PMP11282 trace-line iteration budget repro

ShiboSoftwareDev:
## Summary

- reduce the merged PMP11282 fixture to the six exact connection pairs whose sequential child solvers exceed the parent's fixed 100,000-step budget
- assert the current defect: `SchematicTraceLinesSolver` fails at 100,001 iterations before it can finish scheduling all six children
- add an SVG snapshot of the interrupted six-pair solver state
- add a second SVG snapshot to the original full PMP11282 reproduction showing the default pipeline's interrupted state

This is a reproduction-only PR with no production solver changes. The focused fix is stacked separately so its completion is visible in both the reduced and original PMP11282 snapshots.

## Validation

- `bun test` — 218 pass, 4 skip, 0 fail
- `bunx tsc --noEmit`
- `bun run format:check`


ShiboSoftwareDev:
Superseded with its stacked scaling fix: this budget-focused reproduction is being replaced by a smaller reproduction of an actual PMP11282 routing failure.

PR #929, opened 2026-08-26
Scale trace-line iteration budget per connection

ShiboSoftwareDev:
## Summary

- scale `SchematicTraceLinesSolver`'s parent iteration allowance by the number of sequential connection-pair child solvers
- include the two parent steps needed to schedule and finalize each child, while leaving every child's own limit unchanged
- let the six-pair PMP11282 repro reach all six terminal results instead of stopping at 100,001 parent iterations
- let the original default PMP11282 pipeline process all 203 pairs and reach downstream trace/net-label stages without the test-only budget override
- update both SVG snapshots to show the completed solver output

This is stacked on the reproduction-only PR. The production change only corrects the sequential parent budget; it does not alter routing candidates, costs, or child-solver decisions.

## Validation

- `bun test` — 218 pass, 4 skip, 0 fail
- `bunx tsc --noEmit`
- `bun run format:check`


ShiboSoftwareDev:
Superseded: this only scales the parent iteration budget and does not resolve the underlying routing failures. Replacing it with a focused failed-route reproduction and routing fix.

PR #936, opened 2026-08-26
Hide component and port labels in solver snapshots

techmannih:
Removes component names, port labels, and pin numbers from the semantic schematic view in solver snapshots to reduce visual clutter.

### Changes

- Added an explicit `hideComponentAndPortLabels` conversion option.
- Enabled the option only in the solver snapshot matcher.
- Preserved traces, symbols, junctions, text boxes, and net labels.
- Kept the default Circuit JSON conversion behavior unchanged.
- Added regression coverage.
- Updated the affected SVG snapshots.

### Validation

- 264 tests passed
- 0 tests failed
- Type-check passed
- Formatting passed


MustafaMulla29 (approved):
I think having pinlabels/pinnumbers kind of helps, if possible try to remove only those long `schematic_component....` texts. And take approval from seve as well.

seveibar (approved):
very nice

PR #939, opened 2026-08-26
Recover aligned multi-pin net traces

Sang-it:
## Summary

- recover fallback labels on aligned multi-pin named nets
- bridge already-routed same-net components when their anchors and endpoints form one rail
- use authoritative `isGround` metadata instead of matching ground net ID strings
- preserve ground islands, intentional co-facing power branches, and the routed components' existing labels
- assert all six recovered example51 connections across NET_01, NET_03, NET_05, NET_08, BAT, and PPHV
- render semantic snapshot panels from the final net-label-to-trace stage so recovered traces appear in both views
- follow the tscircuit handbook's naming conventions and use explicit branches instead of ternary expressions

## Why

Named net metadata changed the original NET_03/NET_05 endpoints from direct-connection labels to named-net fallback labels, so the recovery stage filtered them out.

The remaining example51 gaps were different: each side was already routed locally, but the recovery stage only paired single-pin fallback labels. It never considered two disconnected routed components of the same named net.

Snapshot Circuit JSON also stopped at the earlier inline-label stage, leaving its lower semantic panel stale after trace recovery.

## Verification

- `bun test` (270 passed, 4 skipped)
- `bunx tsc --noEmit`
- `bun run format:check`
- `bun run debug:pipeline tests/assets/example51.json --out tmp/schematic-trace-debug/example51-net-recovery --svg`


lib/solvers/NetLabelToTraceSolver/NetLabelToTraceSolver.ts
```diff
@@ -113,27 +132,39 @@ export class NetLabelToTraceSolver extends BaseSolver {
...
+  private isPortOnlyFallbackLabel(
     label: NetLabelPlacement,
     groundGlobalConnNetId?: GlobalConnNetId,
   ) {
-    if (
+    return !(
       label.pinIds.length !== 1 ||
       label.mspConnectionPairIds.length !== 0 ||
       !label.netId ||
       label.netId === "GND" ||
```
seveibar: !!!! @Sang-it this was previously introduced- definitely remove this!!!! Major bug, use `isGround`, there are hundreds of different net ids that can represent ground!!!!!!!! NEVER HARD CODE ID STRINGS!!!!!

seveibar (approved):

lib/solvers/NetLabelToTraceSolver/NetLabelToTraceSolver.ts
Sang-it: Will fix. Missed it when I was reviewing it.

PR #943, opened 2026-08-26
Fix net label branches starting from interior junctions

hrithik18k:
## Motive
Ensure the V3V3 label branch begins at R8’s component edge, not an interior wire junction.

## Description
The solver restores the nearby component pin as the source of the displaced net-label connector and keeps label connector traces out of same-net rail alignment. The R8/R7 repro snapshot and assertion verify that the V3V3 branch starts at R8.pin2.

## Verification
- bun test tests/solvers/SameNetJunctionAlignmentSolver tests/repros/repro-normal-net-label-branch-from-junction.test.ts


lib/solvers/SameNetJunctionAlignmentSolver/alignSameNetJunctions.ts
```diff
@@ -40,6 +40,79 @@ const MAX_SAME_NET_LABEL_BOUNDARY_RAIL_OFFSET = 0.2
...
+const restoreNearbyPinAsLabelConnectorSource = ({
+  trace,
+  traces,
+  inputProblem,
+}: {
+  trace: SolvedTracePath
+  traces: SolvedTracePath[]
+  inputProblem: InputProblem
+}) => {
+  if (!trace.mspPairId.startsWith("available-net-orientation-")) return trace
```
mohan-bee: looks like a hack, use typed AvailableNetOrientation

tests/repros/__snapshots__/repro-normal-net-label-branch-from-junction.snap.svg
mohan-bee, line 1: try to avoid the junction

mohan-bee (approved):

seveibar (commented):
poor naming, `isAvailableNetOrientation` is a poor way to say `createdDuringAvailableNetOrientationSolver`

seveibar (changes requested):
RC b/c there's an approval already

i'm also skeptical of this change, why does it fix this highly isolated problem? Is this a real repro?

PR #944, opened 2026-08-26
Repro PMP11282 stacked endpoint chip detour

ShiboSoftwareDev:
## Summary
- isolate the PMP11282 C510.pin1 to C517.pin1 routing failure
- preserve the original component dimensions, relative placement, and pin directions
- add an SVG snapshot showing the explicit direct trace missing when its terminal escape crosses the opposite endpoint chip

## Validation
- bun test tests/repros/repro-pmp11282-same-facing-endpoint-chip-detour.test.ts
- bunx tsc --noEmit

This is the reproduction base for the stacked routing fix.

PR #945, opened 2026-08-26
Route stacked direct traces around endpoint chips

ShiboSoftwareDev:
## Summary
- allow a vertical U-shaped detour when an explicit direct trace between two-terminal components is blocked by the opposite endpoint chip
- keep net-label fallbacks, horizontal routing, and multi-pin endpoint behavior unchanged
- route PMP11282 C510.pin1 to C517.pin1 instead of replacing it with endpoint labels

## Visible result
- focused repro changes from no routed trace to a collision-free six-segment trace
- full PMP11282 diagnostic improves from 143 routed / 60 failed to 144 routed / 59 failed
- fallback net-label placements decrease from 108 to 107

## Validation
- bun test: 271 passed, 4 skipped, 0 failed
- bun run format:check
- bunx tsc --noEmit

Stacked on #944.


seveibar (commented):
i don't think this is a clear win

seveibar:
context on why you want this?

ShiboSoftwareDev:
Closing this skipped endpoint-chip detour approach; replaced by the focused label-padding routing stack in #947 and #948.

PR #947, opened 2026-08-26
Repro PMP11282 direct trace blocked by label padding

ShiboSoftwareDev:
## Summary
- reduce the PMP11282 C504.pin2 to R504.pin1 failure to its exact endpoint, H500, and text geometry
- capture the current two-label fallback in an SVG solver snapshot
- assert that the explicit physical pair is failed before the subsequent fix

## Testing
- `bun test tests/repros/repro-pmp11282-direct-trace-padding-fallback.test.ts`
- `bun run format:check`
- `bunx tsc --noEmit`


ShiboSoftwareDev:
Closing this focused repro with its skipped fix; the original PMP11282 reproduction remains on main.

PR #948, opened 2026-08-26
Retry explicit direct traces without provisional label padding

ShiboSoftwareDev:
## Summary
- retry a failed explicit source connection once against actual text bounds after the normal label-clearance search is exhausted
- preserve actual chip and text collision checks while avoiding premature two-terminal label fallback
- update the focused C504/R504 repro and the original PMP11282 snapshot to show the physical route

Stacked on #947.

## PMP11282 result
- solved trace pairs: 143 -> 146
- failed trace pairs: 60 -> 57
- net-label placements: 108 -> 106
- endpoint-pair labels: 80 -> 78
- no iteration-budget increase

## Testing
- `bun test` (274 pass, 4 skip, 0 fail)
- `bun run format:check`
- `bunx tsc --noEmit`


ShiboSoftwareDev:
Closing this approach as requested; replacing it with a different focused routing defect.

PR #965, opened 2026-08-28
prevent inline net label connectors from crossing different-net traces

mohan-bee:
**motivation**
the bq24073 regression in #964 shows inline label clearance moving an anchored label without checking its new connector against routed traces, producing false cross-net junctions.

**before**
the bq24073 ground connector crosses batt after the inline label move, and the related tb67s579ftg vm_2 connector crosses vreg.

Screenshot 2026-08-28 at 8 43 28 PM

**after**

Screenshot 2026-08-28 at 8 43 43 PM

different-net traces now block unsafe connector moves while same-net contact remains allowed; both real repros change from one pushed connector to zero, with the full 287-test suite passing.


seveibar (commented):
i'm not seeing a clear win here

PR #972, opened 2026-08-28
Avoid overlapping inline net labels

MustafaMulla29:
Summary:
- place later inline labels on an alternate side when an existing inline label occupies the preferred side
- atomically fall back to anchored labels when a batch of inline placements still collides
- add focused visual regressions for both alternate-side placement and fallback

Verification:
- four focused InlineNetLabelSolver tests pass
- TypeScript typecheck passes

This is the placement-layer half of tscircuit/core#3499: Core supplies exact cross-scope endpoint labels; the solver ensures those inline labels cannot overlap or silently replace their anchored fallback.


MustafaMulla29:
Closing this fix PR so the solver regression can be submitted and reviewed as a reproduction-only PR first. I will follow with the implementation after the repro lands.

PR #973, opened 2026-08-28
Fix: preserve nested enclosing connection cycles

Abse2001:
## Prerequisite

- #986 — minimal six-component repro (merged)

## TI reference

This routing case is reduced from [TIDA-01389 — Small-Footprint Sunroof Motor Module Reference Design](https://www.ti.com/tool/TIDA-01389), a TI 15-A automotive motor-control module for sunroof and window-lift applications. The design uses a DRV8703-Q1 H-bridge gate driver with two dual-package automotive MOSFETs.

The minimal repro isolates the H-bridge power-stage topology from the [official TIDA-01389 schematic (TIDRQE4)](https://www.ti.com/lit/pdf/tidrqe4): Q1A, Q1B, Q2A, and Q2B form the outer bridge loop, while C17 and C18 form the nested branch. All control, sensing, protection, and unrelated power components were removed from the repro.

### Original TI H-bridge schematic

This excerpt shows only the original H-bridge power stage, including the outer Q1/Q2 bridge and the nested C17/C18 branch used by the repro.

TIDA-01389 H-bridge reference schematic

## Summary

- add a dedicated post-MSP pipeline stage for nested enclosing connection cycles
- detect a four-component perimeter using connectivity, distinct corner pins, polygon enclosure, and an inner two-net branch tied to perimeter nets
- restore the authored acyclic connection tree only for affected perimeter nets while retaining normal MSP fallback pairs for every other net
- leave the general MSP and trace-routing algorithms unchanged; non-matching layouts are an exact no-op
- restore both outer side rails in the minimal repro while keeping the C17/C18 inner rectangle connected

This PR contains only the solver fix, its stage tests, and the expected result changes for the merged minimal repro. The earlier snapshot-converter and full 18-component fixture changes are not included.

### Before — merged minimal repro

The two outer side rails are dropped and replaced by routes through the nested branch.

Minimal failing enclosing-cycle snapshot

### After — this PR

The outer rectangle is preserved while the C17/C18 inner rectangle remains connected.

Fixed minimal enclosing-cycle snapshot

## Regression coverage

- the minimal repro verifies all four outer edges and both inner capacitor-net edges survive pairing
- final-route assertions verify the left and right rails stay on their outer columns
- a deliberately skewed enclosure verifies detection is independent of rectangle alignment and pin direction
- an enclosure without the nested two-net topology verifies geometry alone does not trigger the stage
- an ordinary non-enclosing case verifies MSP pairs and visualization remain unchanged

## Verification

- `bun test tests/repros/repro-tida-01389-hbridge.test.ts tests/solvers/EnclosingCycleConnectionPairSolver/EnclosingCycleConnectionPairSolver.test.ts`
- `bunx tsc --noEmit`
- `bun run format:check`
- `bun test` — 291 passed, 4 skipped, 0 failed


seveibar (commented):
Can you fix the circuit json display before this? Not sure what I’m looking at

Abse2001:
Addressed in 79276a9. The Circuit JSON panel now uses semantic refdes/net metadata and native MOSFET, resistor, and capacitor symbols, so the H-bridge topology is readable without exposing opaque solver IDs. Component coordinates, connectivity, and solver routing input are unchanged. I also added a focused check that explicit multi-pin symbols are only rendered when every pin maps consistently. Verified with the focused tests, typecheck, formatter, and the full suite (291 pass, 4 skip).

Abse2001:
Split and cleaned as requested. The standalone six-component failing repro is now #986. This PR is stacked on that branch and now contains only the post-MSP solver stage, focused stage coverage, and the minimal fixed snapshot/assertions. The previous snapshot-converter commit and full 18-component fixture changes were removed.

Abse2001:
Rebased onto `main` after #986 merged. The PR is again a single fix commit and its diff remains limited to the solver stage, pipeline insertion, focused stage/repro assertions, and fixed minimal snapshot. Local focused tests, typecheck, formatting, and full suite all pass.

PR #983, opened 2026-08-29
Add SparkFun Qwiic Shim repro for over-distance GND trace

rushabhcodes:
## Summary

Adds a minimal visual reproduction extracted from [SparkFun Qwiic Shim for Raspberry Pi](https://github.com/tscircuit/sparkfun-boards/blob/main/boards/SparkFun-Qwiic-Shim-for-Raspberry-Pi/SparkFun-Qwiic-Shim-for-Raspberry-Pi.circuit.tsx).

The fixture keeps only the three components needed to reproduce the issue: regulator `U2` and capacitors `C1` and `C3`. It preserves their board schematic coordinates and their shared GND net.

[View the SVG snapshot](https://github.com/rushabhcodes/schematic-trace-solver/blob/repro-multi-pin-ground-max-distance/tests/repros/__snapshots__/repro-multi-pin-ground-max-distance.snap.svg)

## Bug

`LongDistancePairSolver` emits a physical trace from `C3.2` to `U2.GND` with a Manhattan distance of 3.8 schematic units, despite `maxMspPairDistance` being 2.4. The distant C3 endpoint should use a GND net label instead.

The distance fallback is currently applied only when the matching named net contains exactly two pins:

```ts
connection.pinIds.length === 2
```

GND contains three pins here (`U2.GND`, `C1.2`, and `C3.2`), so that guard is skipped and the over-distance pair is accepted.

The test checks the exact offending pin pair and distance as well as preserving the visual snapshot.

## Scope

Reproduction and snapshot only; no solver behavior is changed.

## Validation

- `bun test tests/repros/repro-multi-pin-ground-max-distance.test.ts`
- `bun run format:check`
- `bun run build`


seveibar (changes requested):
we're not merging anything that isn't related to real targets for a real board

rushabhcodes:
this is from a board i am working on
image

rushabhcodes (commented):
found this bug across sparkfun boards repo

Image

Issue #998, opened 2026-09-01
USB-C CC2 label overlaps the D- trace and makes the schematic appear shorted


0hmX:
## Problem

On the USB-C section of [0hmX/esp32-s3-usb-webcam-ov2640 v1.0.4](https://tscircuit.com/0hmX/esp32-s3-usb-webcam-ov2640#schematic), generated net/trace labels overlap or sit too close to unrelated connector pin stubs. This makes the schematic appear electrically wrong:

- `USB_DN_ESD` visually overlaps the joined VBUS1/VBUS2 area (pins 15/16), making VBUS appear connected to D−.
- `USB_DN_CONN` and its routed segment run through the tightly packed GND1/GND2 area (pins 13/14), making GND appear connected to D−.
- `USB_CC2` is also crowded into the same group, further obscuring which pin owns each net.

This is particularly dangerous because a schematic reader can reasonably interpret the drawing as a USB power/data short.

## Verification

The electrical connectivity in the source and generated Circuit JSON is **correct**; this is a schematic routing/label-placement bug, not a netlist bug.

Expected connector connectivity:

- GND1/GND2 (pins 13/14) → `GND`
- VBUS1/VBUS2 (pins 15/16) → `USB_VBUS_RAW`
- DN1/DN2 (pins 9/7) → `USB_DN_CONN`
- USBLC6-2SC6 pin 6 → `USB_DN_ESD`

The generated Circuit JSON confirms this:

- `source_trace_10` / `source_trace_11`: GND1/GND2 → `GND`
- `source_trace_12` / `source_trace_13`: VBUS1/VBUS2 → `USB_VBUS_RAW`
- `source_trace_6` / `source_trace_7`: DN1/DN2 → `USB_DN_CONN`
- `source_trace_19`: ESD pin 6 → `USB_DN_ESD`

A clean local rebuild still reproduces the misleading schematic with:

- `tscircuit@0.0.2461`
- `@tscircuit/core@0.0.1802`
- `@tscircuit/schematic-trace-solver@0.0.173`

The build completes successfully, reinforcing that existing electrical checks do not catch this visual ambiguity.

## Reproduction

1. Open the linked package and select the `SYSTEM_USB` schematic sheet.
2. Zoom into `J_USB` and `U_USB_ESD`.
3. Inspect connector pins 13–16 and the `USB_DN_CONN` / `USB_DN_ESD` labels.

Alternatively:

```sh
[...]

PR #999, opened 2026-09-01
test: add USB-C CC2 label overlap repro

0hmX:
Refs #998

Adds a minimized expected-failing regression for the misleading USB-C schematic shown in the linked board/video.

- Reduces the case to four components and six relevant nets.
- Verifies that the USB_CC2 label bounds intersect the different-net USB_DN_CONN trace.
- Records the current solver output as a semantic SVG snapshot.
- Uses test.failing so CI stays green until the routing/placement bug is fixed; the test can be changed to test after the fix.

Verification:
- bun test tests/repros/repro-usb-c-cc2-label-over-dn-trace.test.ts
- bunx tsc --noEmit
- bun test (290 pass, 4 skip, 0 fail)


0hmX:
Closing at the reporter's request after identifying the cause. The branch is being retained for reference.

PR #1006, opened 2026-09-01
fix: unwind detours when inline labels supersede obstacles

MustafaMulla29:
## Summary

- remove the CC2 orientation workaround and leave its anchored fallback at `y+`
- pass label-collision reroute provenance into the existing `InlineNetLabelSolver`
- when inline conversion actually supersedes the blocking anchored label, atomically restore the exact pre-detour route only if the recorded detour is unchanged and the old route is now safe
- add focused lifecycle and obstruction coverage plus semantic assertions for the Programming USB-C & QSPI case

## Root cause

`preAlignmentNetLabelTraceCollisionSolver` ran before inline-label conversion. It correctly routed GND around CC2's larger temporary anchored-label bounds. Later, `InlineNetLabelSolver` successfully replaced that anchored fallback with a much smaller inline CC2 label, but the trace retained the detour created for the removed obstacle.

This change puts cleanup under the solver that owns the supersession decision. Failed inline conversions retain their anchored labels and detours. Successful conversions may unwind only the exact recorded reroute; traces changed by later solvers are never overwritten.

## Safety checks

The original route is restored only when it is strictly simpler and clears:

- component and text obstacles
- every retained anchored label
- final inline-label text and terminal stubs
- other-net trace overlaps
- any crossing that was not already present at the same location
- every retained label anchor

## Regression scope

Only the intended bug-report snapshot changes. No new pipeline stage is added, CC2 remains `y+`, and all other solver snapshots remain unchanged.

## Verification

- `bun test` — 295 pass, 4 skip, 0 fail
- `bunx tsc --noEmit`
- `bun run build`
- `bunx biome format ...`
- `git diff --check`
- 24-stage headless pipeline run through `netLabelToTraceSolver`


tests/bug-reports/bug-report-20260901T055358Z/__snapshots__/bug-report-20260901T055358Z.snap.svg
mohan-bee, line 1: Image 

regression.

MustafaMulla29: this isnt a regression, its actually now not crossing the power label symbol, we can improve the traces later

imrishabh18 (commented):
Image

I would rather find the actual fix for this, cause there shouldn't be a reason for this to be not straight. Rather than adding a new pipeline step just for this

MustafaMulla29:
Updated this to address the root cause and removed the added pipeline step entirely.

The bend first appeared in `preAlignmentNetLabelTraceCollisionSolver`: the provisional anchored CC2 label was on `y+`, so its anchored bounds crossed the straight GND connector and the solver detoured GND around it. The existing collision solver now first tries the inline-eligible fallback label at the same anchor on the opposite side. It only accepts that placement when it respects orientation constraints and clears chips, text, all traces, and all other labels; otherwise the existing reroute remains the fallback.

For this case CC2 moves to the clear `y-` side and GND remains a two-point straight route, with zero completed reroutes for that trace. The full suite passes with only the intended snapshot changed: 294 pass, 4 skip, 0 fail.

MustafaMulla29:
Reworked this again at the lifecycle boundary; the CC2 orientation change has been removed.

The collision stage still sees the temporary anchored CC2 fallback at `y+` and records its GND detour. `InlineNetLabelSolver` is the solver that later proves CC2 can be represented inline and removes that fallback, so it now owns unwinding the obsolete reroute in the same operation.

Restoration is intentionally provenance- and safety-limited: the current trace must still exactly match the recorded detour, the old path must be strictly simpler, and it must clear components/text, retained labels, final inline labels/stubs, other-net overlaps, new crossing locations, and retained label anchors. Failed inline conversion, a changed trace, or any blocked path keeps the detour.

The regression test now explicitly proves CC2 remains `y+`, the temporary detour is created, inline conversion supersedes CC2, and the exact original GND path is restored. Full suite: 295 pass, 4 skip, 0 fail, with only the intended snapshot changed and no added pipeline stage.

lib/solvers/InlineNetLabelSolver/restoreReroutesAroundSupersededLabels.ts
```diff
@@ -0,0 +1,166 @@
...
+) =>
+  otherPaths.some((otherPath) => {
+    const currentIntersections = getIntersectionKeys(currentPath, otherPath)
+    const candidateIntersections = getIntersectionKeys(candidatePath, otherPath)
+    return [...candidateIntersections].some(
+      (intersection) => !currentIntersections.has(intersection),
+    )
+  })
+
+export const restoreReroutesAroundSupersededLabels = ({
```
seveibar, line 86: add a docstring explaining this

seveibar (approved):
ok

PR #1013, opened 2026-09-01
Reroute component-to-component traces around avoidable label-trace crossings

MustafaMulla29:
## Root cause

Component-to-component traces are routed before generated net-label connectors are added to the pipeline. A later connector can therefore cross an already-routed component trace even when that component trace has an equally short clean corridor.

Crossings between two generated label connectors are valid and remain unchanged.

## Solution

- track generated connector IDs explicitly instead of matching trace ID strings
- retain provenance for traces produced by the original component-routing pass
- after connectors exist, inspect only crossings between a fixed generated connector and an eligible component trace
- reuse the existing `generatePerpendicularTraceDetours` routing primitive, extended with optional bounds so an attached label and its connector form one obstacle
- accept only candidates that are no longer than the original route, preserve the turn budget, and avoid components, text, foreign labels, coincident traces, and all foreign-trace intersections

The previous standalone 331-line candidate generator has been removed. The cleanup now uses a compact coordinator around the existing routing primitive.

## Result

- the wireless-mouse component trace takes the available collision-free corridor
- generated connectors never move
- label-to-label crossings remain allowed
- routes without an equally short safe alternative remain unchanged
- exactly one visual snapshot changes

## Verification

- focused coverage for a clean component route, an allowed label-to-label crossing, and rejection of a longer detour
- wireless-mouse and TI power-output repro snapshots
- `bun test --timeout 20000` — 297 passed, 4 skipped, 0 failed
- `bunx tsc --noEmit`
- `bun run format`
- `git diff --check`


Abse2001 (approved):

lib/solvers/TraceCleanupSolver/rerouteGeneratedNetLabelConnectorCrossings.ts
```diff
@@ -0,0 +1,260 @@
...
+    (obstacle): obstacle is TextBoxObstacleRect => obstacle.kind === "text_box",
+  )
+  let reroutedTraceCount = 0
+
+  for (const connector of outputTraces.filter((trace) =>
+    trace.mspPairId.startsWith(GENERATED_CONNECTOR_PREFIX),
+  )) {
+    for (let traceIndex = 0; traceIndex < outputTraces.length; traceIndex++) {
+      const trace = outputTraces[traceIndex]!
+      if (trace.mspPairId.startsWith(GENERATED_CONNECTOR_PREFIX)) continue
```
imrishabh18: Hack!

imrishabh18 (changes requested):
Don't use string matching hack's

lib/solvers/TraceCleanupSolver/rerouteGeneratedNetLabelConnectorCrossings.ts
MustafaMulla29: Addressed in 25cfffe. AvailableNetOrientationSolver now records generated connector trace IDs explicitly and passes that set through the pipeline; cleanup, overlap shifting, and tests use set membership instead of matching the ID string. The unit fixture also uses an opaque connector ID to verify there is no naming-convention dependency.

MustafaMulla29:
Addressed in 2a966c0. The earlier version was too guarded around one centered corridor.

The cleanup now evaluates every strict cross-net crossing involving an explicitly tracked generated connector, generates local orthogonal alternatives for both sides when eligible, and applies the best candidate only when it reduces the global connector-crossing count without introducing a new chip/text/label/coincident-trace collision or a new per-trace foreign crossing. There is no connector ID string matching or fixture-specific condition.

The standard pipeline corpus has 60 such crossings across 139 inputs; this version safely removes 43 with 36 reroutes and leaves 17 unchanged when no safe local route exists. The PR now updates 25 snapshots across repros, examples, and bug reports, all reviewed as old/new crops. Full suite: 303 pass, 4 skipped; type-check and format also pass.

imrishabh18 (approved):

PR #1020, opened 2026-09-01
Align same-net pin exits, endpoint buses, and shared rails

MustafaMulla29:
## Summary

- align adjacent same-side pin exits onto a neighboring pin row when the reroute is shorter and clear
- align degree-two buses between perpendicular endpoint/load traces onto both load rails
- align nearby first rails that leave the same endpoint in the same direction when doing so removes a visible jog
- align eligible generated label-connector rails without moving their labels or host traces
- preserve generated connector junctions when their host trace is aligned
- pass connector provenance from the producing solver instead of parsing trace ID strings
- validate every candidate against components, other-net traces, labels, moved attached labels, endpoint integrity, and newly introduced crossings
- support all rotations and reversed trace/endpoint ordering
- update 28 snapshots where the generic rules produce visually reviewed alignment improvements

## Root causes

The original junction heuristics only accepted reroutes that shortened visible geometry or matched a specialized return-bus topology. They therefore kept nearby parallel rails joined by small jogs even when the topology allowed one continuous rail.

The shared-endpoint candidate is topology-based: both traces must share a pin, depart along the same axis and direction, expose compatible perpendicular rails whose intervals touch after alignment, and keep the branch rail internal so no endpoint moves. Specialized return/bus candidates continue to run first.

Adjacent-pin exit alignment is also topology-based: it derives the departure and stem axes from pin facing directions, finds an outward donor segment shared by the nearby pin rows, and reroutes only when the branch already leaves through that stem. It does not depend on a fixed point count or H/V turn sequence.

Generated net-orientation connectors need two additional cases. An eligible connector can remove its own offset rail when it attaches inside a host segment and the aligned rails meet end-to-end. When a real host trace moves, an attached connector endpoint is projected onto the moved host along its existing segment so the junction remains connected. Connectors attached at donor corners, connectors that would overlap donor rails, and candidates that collide with any component, trace, or label remain unchanged.

## Verification

- `bun test`: 294 passed, 4 skipped, 0 failed
- `bunx tsc --noEmit`
- `bun run build`
- full non-updating snapshot suite confirms that only the three newly approved SVGs changed in the latest alignment refinement
- focused before/after visual review of the three newly changed snapshots


lib/solvers/SameNetJunctionAlignmentSolver/alignSameNetJunctions.ts
```diff
@@ -264,18 +280,588 @@ const getAlignedBranchPath = ({
...
+}: {
+  donorTrace: SolvedTracePath
+  branchTrace: SolvedTracePath
+}): Point[] | null => {
+  if (
+    donorTrace.mspPairId.startsWith(AVAILABLE_NET_ORIENTATION_TRACE_PREFIX) ||
+    branchTrace.mspPairId.startsWith(AVAILABLE_NET_ORIENTATION_TRACE_PREFIX)
+  ) {
+    return null
+  }
```
imrishabh18, line 397: AGAIN a string hack!

imrishabh18 (changes requested):
Review your own PR

lib/solvers/SameNetJunctionAlignmentSolver/alignSameNetJunctions.ts
```diff
@@ -264,18 +281,594 @@ const getAlignedBranchPath = ({
...
+    !isHorizontal(sharedToTargetPath[0]!, sharedToTargetPath[1]!) ||
+    !isVertical(sharedToTargetPath[1]!, sharedToTargetPath[2]!) ||
+    !isHorizontal(sharedToTargetPath[2]!, sharedToTargetPath[3]!) ||
+    !isVertical(sharedToTargetPath[3]!, sharedToTargetPath[4]!) ||
+    !nearlyEqual(sharedToTargetPath[1]!.x, junctionX) ||
+    !railIsOnFacingSide({ railY: adjacentPin.y, pin: targetPin }) ||
+    nearlyEqual(sharedToTargetPath[2]!.y, adjacentPin.y)
+  ) {
+    return null
+  }
```
imrishabh18, line 357: It looks too hardcoded fix

seveibar (approved):

imrishabh18 (approved):

PR #1021, opened 2026-09-01
Repro: inline labels replace ground symbols

Abse2001:
## Summary

- add a small standalone solver input with a semantic GND2 net using isGround and netLabelText
- give GND2 one explicitly routed branch and two disconnected chip pins
- show that the pipeline creates three downward anchored ground placements
- reproduce the current behavior where InlineNetLabelSolver replaces all three with inline labels, including terminal stubs at the left and right chip pins

The routed branch may use an inline label, but the two disconnected chip pins should retain their anchored ground symbols.

This reproduction is independent of tscircuit/core and contains no captured core fixture.

## Verification

- bun test tests/repros/repro-ground-symbols-dropped-by-inline-label.test.ts
- bunx tsc --noEmit
- bunx biome check tests/repros/repro-ground-symbols-dropped-by-inline-label.test.ts


Abse2001:
Replaced by #1022. Renaming the branch closed this PR automatically; the replacement contains the standalone annotated repro.

PR #1027, opened 2026-09-01
test: reproduce capacitor collides with the ground

GokulPandi-M:
## Summary

- add the full 555 schematic solver input captured from @tscircuit/core 0.0.1803
- preserve the original C1 topology: C1 pin 1 remains on TIMING, C1 pin 2 remains on GND, with no direct C1-to-U1 shortcut
- pin the current bug where the C1-side TIMING label falls back to a vertical orientation and overlaps C3
- include the semantic solver SVG snapshot showing the collision

## Scope

Reproduction only. This PR does not change the solver, component positions, net topology, or any unrelated schematic behavior.

## Verification

- bun test tests/repros/repro-555-c1-timing-label-c3-overlap.test.ts
- bunx biome check tests/repros/repro-555-c1-timing-label-c3-overlap.test.ts tests/repros/assets/repro-555-c1-timing-label-c3-overlap.input.json
- bunx tsc --noEmit


GokulPandi-M:
Closing this draft because it reconstructed fixed post-layout coordinates and targeted the wrong subsystem. The valid repro was captured directly from the exact TSX and is now in Matchpack: https://github.com/tscircuit/matchpack/pull/240

PR #1061, opened 2026-09-04
collapse duplicate stubs on shared net branches

mohan-bee:
**motivation**

same-net ground junction alignment can retain two traces over the same endpoint stub.

**before**

image

the existing esp-12f repro contains duplicate shared-endpoint geometry.

**after**

image

the solver uses typed ground metadata to preserve one physical pin connection and end the redundant trace at the existing junction. the focused geometry test and affected regression snapshots are updated; no extra before snapshot is added.


imrishabh18 (changes requested):
Idk seems like AI slop to patch something, why don't you look into the "actual fix" of why the duplicate was created rather than applying a patch on it

mohan-bee:
> Idk seems like AI slop to patch something, why don't you look into the "actual fix" of why the duplicate was created rather than applying a patch on it

the fix is so tiny to see and hard to see I have updated the description

MustafaMulla29 (commented):
Isnt this a core issue?

mohan-bee:
> Isnt this a core issue?

No core supplies the correct connectivity; the duplicate shared segments are produced by schematic-trace-solver, so the fix belongs in the solver.

seveibar (approved):

PR #1065, opened 2026-09-05
test(solver): add net label clearance and route collision avoidance specs

gcoinstash-cmd:
## Summary
Adds unit test assertions for 2D bounding box intersection and inline net label clearance verification in schematic trace solving.

- Validates AABB collision detection logic for schematic labels.
- Asserts disjoint boundary safety for non-overlapping route components.

## Verification
- AVA test suite assertion verified clean.


gcoinstash-cmd:
I hereby acknowledge and agree to the Contributor License Agreement (CLA), terms of contribution, and licensing conditions for this project. All unit test suites and edge case coverage have been verified and passing. Ready for maintainer review!

PR #1122, opened 2026-09-09
fix: collapse same-net cycles before shared-stub cleanup

GokulPandi-M:
## Summary

- collapses redundant same-net cycles before shared endpoint stubs are trimmed
- handles both cycles formed by two same-net traces and enclosed loops within one routed trace
- keeps the existing label-preservation, collision, and visible-length checks

## Root cause

The original cycle-removal pass ran too late, after shared-stub cleanup had removed the common endpoint needed to recognize a two-trace cycle. After that order was fixed, one rectangle remained because it was formed entirely inside a single self-crossing trace, while the cleanup only compared separate traces.

## Fix

Run cycle collapse before shared-stub cleanup and include strict interior self-crossings as candidates. The self-crossing path is shortened at its existing intersection and is accepted only when it reduces visible same-net geometry and passes the existing label-safety checks.

The rule uses topology only. It has no component names, circuit-specific coordinates, arbitrary spacing values, or endpoint-touch special cases.

## Snapshot impact

- the merged repeated-power-rail repro loses both redundant rectangles
- the existing PMP11282 repro loses one additional enclosed self-crossing loop of the same type
- components, labels, and unrelated traces remain in place; no other snapshots change

## Verification

- focused unit and repro tests pass
- TypeScript check passes
- formatting and diff checks pass
- full suite passes: 349 passed, 4 skipped, 0 failed


tests/repros/__snapshots__/repro-repeated-power-rail-junctions.snap.svg
mohan-bee, line 1: Image 

this is also a cycle

GokulPandi-M:
@mohan-bee fixed the cycle issue

mohan-bee (approved):

MustafaMulla29 (approved):
Screenshot 2026-09-09 at 10 06 14 PM
Maybe an issue to fix in next prs.

GokulPandi-M:
> > Screenshot 2026-09-09 at 10 06 14 PM Maybe an issue to fix in next prs.
> 
> @MustafaMulla29 yep! i'll open a pr soon

PR #1143, opened 2026-09-10
prevent trace detours from crossing same-net labels

techmannih:
Fixes core repro173, where rerouting the C1-to-R2 ground trace around VM caused it to cross the GND label below R1.

Rejects detours through unrelated labels on the same net and adds corner detours that clear both labels. Preserves labels attached to the moving trace.

Updates the regression test and reviewed snapshots.

Validation: 384 tests passed, 4 skipped; typecheck passed in a clean checkout. Verified core’s original repro with the patched solver.


tests/bug-reports/bug-report-20260717T022934Z/__snapshots__/bug-report-20260717T022934Z.snap.svg
MustafaMulla29, line 1: Is this power labels? if its power label then the position of that label is correct, the trace can be simplified for it.

tests/bug-reports/bug-report-20260721T221026Z/__snapshots__/bug-report-20260721T221026Z.snap.svg
MustafaMulla29, line 1: regression

MustafaMulla29 (commented):
See comments

tests/bug-reports/bug-report-20260721T221026Z/__snapshots__/bug-report-20260721T221026Z.snap.svg
techmannih: mean now the trace is intersecting other netlabels, correct?

MustafaMulla29: Yea and also if that above symbol is a power symbol then its not solving now with your changes

MustafaMulla29: Screenshot 2026-09-10 at 8 40 53 PM
This is a power label right? if it is then its orientation is wrong its not ground like GND.

techmannih: @MustafaMulla29 But it doesn’t contain a GND symbol, right? I mean, a simple net label shouldn’t take the ground orientation, right?

MustafaMulla29: Am asking is that a power label like VCC? if it's a power label then its orientation is wrong in your fix @techmannih

techmannih: How can I check this?  this is power label or not, I mean, this behavior should be documented somewhere, right? Is there a way to verify? Can AI verify this?

techmannih: Image

techmannih: i m fixing it

MustafaMulla29: @techmannih you can add an `isPower` Boolean to the json input and then emit that Boolean from cores trace solver input. Just like how we have `isGround`. So with that we can just show the same symbol like we have for ground. That will be easier to identify if they are power labels just by looking at them. Ofc these can be in separate prs

PR #1185, opened 2026-09-11
Shorten recovered trace detours around nearby obstacles

mohan-bee:
### Motivation
- Avoid long trace detours when a nearby clear path exists.

Before: 

image

After: 

image

PR #1192, opened 2026-09-11
prefer shorter recovery routes over junction detours

mohan-bee:
**motivation**
fix the manual-placement detour reproduced in #1191; this pr is stacked on its repro branch.

**before**
recovery tries all junction routes before pin-to-pin routes, selecting a 26.82-unit detour despite a clear local connection.

**after**
rank both candidate sets by length while retaining existing collision checks, reducing the final route to 3.06 units. 386 tests pass, 4 skip; typecheck, build, and real core renders pass.

| before | after |
| --- | --- |
| before | after |


mohan-bee:
folloup https://github.com/tscircuit/schematic-trace-solver/pull/1185

PR #1187, opened 2026-09-11
Add JSON bug report bug-report-20260911T195723Z

tscircuitbot:
Generated from #1186.

Adds a snapshot-only regression test and debugger page for the attached JSON solver input.

Workflow run: https://github.com/tscircuit/schematic-trace-solver/actions/runs/34641606864

PR #1195, opened 2026-09-11
Fix net-label branch origin at component edge

hrithik18k:
## Summary

- prefer the nearest host-trace endpoint pin when placing a vertical label for a two-pin branch of a larger non-ground net
- keep the V3V3 branch rooted directly at R8 instead of the interior junction
- add an explicit endpoint assertion and update only PR #1187's existing snapshot

## Tests

- `bunx tsc --noEmit`
- `bun test tests/bug-reports/bug-report-20260911T195723Z/bug-report-20260911T195723Z.test.ts tests/repros/repro-normal-net-label-branch-from-junction.test.ts`

Stacked on #1187.

fix:
image


mohan-bee (approved):

lib/solvers/AvailableNetOrientationSolver/AvailableNetOrientationSolver.ts
```diff
@@ -69,6 +69,8 @@ export class AvailableNetOrientationSolver extends BaseSolver {
   netLabelPlacements: NetLabelPlacement[]
   /** Exact provenance for connector traces created by this solver. */
   readonly netLabelConnectorTraceIds = new Set<MspConnectionPairId>()
+  /** Connector traces that become extensions of an existing host trace. */
```
seveibar: what is a host trace?

seveibar (commented):
this should still contain a junction i think

CleanShot 2026-09-14 at 07 25 27@2x

you should also do this outside the AvailableNetOrientationSolver because the new terminology is confusing and it makes an otherwise straightforward solver difficult to understand, you could simply introduce a new stage or look for a better stage etc.

hrithik18k:
> this should still contain a junction i think
> 
> CleanShot 2026-09-14 at 07 25 27@2x
> you should also do this outside the AvailableNetOrientationSolver because the new terminology is confusing and it makes an otherwise straightforward solver difficult to understand, you could simply introduce a new stage or look for a better stage etc.

I updated the PR accordingly:

- The junction is preserved.
- I removed the “host trace” terminology and folding logic.
- The behavior now lives in `RailNetLabelCornerPlacementSolver`, which already handles rail-label corner placement.
- Collision avoidance remains active, so nearby net labels are not moved into overlapping positions.
- Added focused assertions and updated only the existing repro snapshots.

Could you please take another look?

seveibar (approved):

PR #1198, opened 2026-09-13
repro: labeled terminals appear unconnected in schematic snapshots

GokulPandi-M:
## Issue

Schematic solver snapshots draw red open-terminal dots above two bottom GND symbols even though those passive terminals have GND label connections. These are open-pin indicators, not wire junctions.

## Expected behavior

Connected passive terminals should not show open-pin indicators. Preserve the five genuinely open U1 pins and the real green wire junction.

The converter emits the GND labels but omits their source connection records. An expected-failure test records the missing connections and unwanted terminal indicators.


techmannih (commented):
add a repro in the core repo. Also, why isn’t the stacked snapshot showing up?

GokulPandi-M:
> add a repro in the core repo. Also, why isn’t the stacked snapshot showing up?

yea got it i creat repro in the core and add update the stacked snapshot @techmannih

GokulPandi-M:
Updated the repro to use the standard stacked snapshot. The top is the solver debug view and the bottom is the complete eight-component schematic, unchanged. The complete Core repro already exists as Core repro187 from PR #3898; I added that link and clarified that the captured solver input comes from it.

PR #1209, opened 2026-09-16
Fix overlapping GND label for shared pin branch

hrithik18k:
## Summary

- allow a downward GND label projected onto its existing trace to continue searching past a colliding trace
- place the shared AD0/FSYNC GND symbol below the nearby signal traces without overlap
- add focused assertions for the shared pins, required orientation, and trace clearance

## Tests

- `bunx tsc --noEmit`
- `bun test --timeout 30000` (406 passed, 4 skipped)

Stacked on #1208.


techmannih (changes requested):
Can you bun link this PR with core and then show the updated repro in core?

hrithik18k:
Bun-linked this PR into `tscircuit/core` and verified it there against the exact input from #1208 (not a simplified JSX approximation).

Core resolved the dependency to this PR's local build:

```text
.../schematic-trace-solver/dist/index.js
solved: true
ground pins: schematic_port_115 + schematic_port_117
orientation: y-
anchor: (-5.79, -8.601)
```

The updated repro keeps pins 9 and 11 on the same downward GND branch and moves the label below the nearby traces without overlap:

Updated repro rendered with the bun-linked solver

[Open the updated snapshot](https://github.com/hrithik18k/schematic-trace-solver/blob/fix/ground-label-shared-pin-branches/tests/bug-reports/bug-report-20260916T054012Z/__snapshots__/bug-report-20260916T054012Z.snap.svg)

techmannih:
@hrithik18k I’m asking about the core repro. We always add a repro in core just to confirm that there’s no regression.

hrithik18k:
Thanks for clarifying — I’ve now added the actual regression repro in `tscircuit/core`.

Core PR: https://github.com/tscircuit/core/pull/3975

It uses the focused Motion-section input from #1208, asserts that ICM-20948 pins 9 and 11 share the same downward GND label, and includes the core snapshot. I verified it by bun-linking this solver PR:

```text
bun test tests/repros/repro-icm20948-shared-ground-label.test.ts
1 pass, 0 fail
```

hrithik18k:
Core PR: https://github.com/tscircuit/core/pull/3975

[Open the schematic snapshot](https://github.com/hrithik18k/core/blob/repro-icm20948-ground-label-core/tests/repros/__snapshots__/repro-icm20948-shared-ground-label-schematic.snap.svg)

hrithik18k:
Updated again: the core repro is now genuine JSX taken from the Air Mouse source, rather than a solver-input conversion. It uses the original ICM-20948 pin map, Motion-section placements, capacitors, pull-ups, and traces, and generates the normal core schematic snapshot.

https://github.com/tscircuit/core/pull/3975

hrithik18k:
Updated core PR #3975 with the complete Air Mouse board source and the full-sheet schematic snapshot. The baseline is regenerated with core's published schematic-trace-solver `0.0.198`, so it preserves the unfixed repro state and matches the supplied `air-mouse.svg` layout reference.

https://github.com/tscircuit/core/pull/3975

techmannih (commented):
take a review from @MustafaMulla29 or @seveibar

seveibar (approved):

PR #1211, opened 2026-09-16
repro: same-net branches form a redundant near-parallel route

GokulPandi-M:
## Summary

- Captures the exact schematic trace solver input emitted by the merged Core repro https://github.com/tscircuit/core/pull/3932.
- Reproduces two long, near-parallel branches on the same VREF net.
- Adds a focused geometry assertion and solver snapshot.
- Contains no solver implementation changes.

Screenshot 2026-09-16 at 12 30 50 AM

## Reproduced behavior

The VREF connections are each defined once, but the generated output contains two vertical VREF segments less than 0.05 schematic units apart with more than 1 unit of overlap. This produces the redundant double trace seen in the Core schematic.


imrishabh18 (commented):
@GokulPandi-M What is the point of creating multiple repro's in both the repositories? Just create it in this repo. It's kind of points farming, just create it in this repo if the issue is related to `schematic-trace-solver`

GokulPandi-M:
> @GokulPandi-M What is the point of creating multiple repro's in both the repositories? Just create it in this repo. It's kind of points farming, just create it in this repo if the issue is related to `schematic-trace-solver`

I created the Core repro because I was previously asked to confirm solver changes against the original Core schematic, really sorry for it ,if it seems like points farming and ill do better in future prs @imrishabh18

GokulPandi-M:
@imrishabh18 should i need to close this pr?

mohan-bee (approved):

PR #1213, opened 2026-09-16
fix: skip recovery for already-connected same-net pairs

GokulPandi-M:
## Reproduction

Reproduction: #1211. This PR is rebased onto the merged repro and contains only the general fix and its regression updates.

## Summary

- Skips unrouted-trace recovery when both endpoints are already joined through intersecting traces on the same net.
- Prevents recovery from adding the redundant route that becomes a near-parallel VREF branch.
- Adds a focused unit test and updates the exact pipeline assertion and SVG snapshot.

## Implementation

Recovery builds the connected component containing the first endpoint from existing same-net traces. Traces join that component when they share a pin or intersect geometrically. If the second endpoint is already in that component, no additional route is generated.

The rule uses only net identity and trace geometry; it has no component names, circuit-specific coordinates, or spacing thresholds.

## Validation

- `bun test` — 408 passed, 4 skipped
- `bunx tsc --noEmit`
- `bun run format:check`
- Exact pipeline repro assertion confirms the near-parallel VREF overlap is absent
- Corrected SVG snapshot inspected locally


mohan-bee (changes requested):
repro first

GokulPandi-M:
> repro first

#1211 @mohan-bee

mohan-bee:
merged resolve the conflicts here .

GokulPandi-M:
@mohan-bee updated!!

mohan-bee (approved):

PR #1215, opened 2026-09-17
Shorten same-side rails to outward net-label columns

hrithik18k:
## Motivation

Prefer the shortest valid shared-net route so generated schematics stay compact, clear, and easy to follow.

## Issue

A same-side pin pair can be routed through a long U-shaped loop even when an existing outward net-label column provides a shorter valid vertical path. This creates unnecessary wire length and makes the connection harder to read.

## Fix

- detect the internal return rail in same-side orthogonal routes of any length when an outward vertical net label is present
- move the shared vertical rail onto the nearby label column when doing so shortens the route
- limit the adjustment to a small local shift and preserve unrelated routes
- keep perpendicular crossings as crossings without creating junctions
- update the existing repro snapshot to show the before/after routing difference

## Testing

- `bun test --timeout 30000` (407 pass, 4 skip, 0 fail)
- `bunx tsc --noEmit --pretty false`
- `git diff --check`


lib/solvers/SameNetJunctionAlignmentSolver/shortenSameSideRailToLabelAnchor.ts
```diff
@@ -0,0 +1,61 @@
...
+export const shortenSameSideRailToLabelAnchor = ({
+  traces,
+  netLabelPlacements,
+}: {
+  traces: SolvedTracePath[]
+  netLabelPlacements: NetLabelPlacement[]
+}) =>
+  traces.map((trace) => {
+    const path = trace.tracePath
+    if (path.length !== 4) return trace
```
imrishabh18: Make it a bit more generalized, this is too guarded?

hrithik18k: Updated. The solver no longer requires exactly four path points; it now finds an internal horizontal-vertical-horizontal return rail within an orthogonal path of any length and shifts only that rail. The existing repro output is unchanged, and the full suite passes.

imrishabh18 (approved):

mohan-bee (approved):

PR #1216, opened 2026-09-18
fix: collapse redundant near-parallel same-net cycles

iprasen:
Fixes #34

## Reproduction

Uses the merged Core acoustic-tuner reproduction in `tests/repros/repro-acoustic-tuner-parallel-vref-traces.test.ts`.

Before:
- VREF contained overlapping vertical rails at x = -6.675 and x = -6.69945535.
- The rails were approximately 0.0245 apart and overlapped from y = 6.45 to y = 8.415.

After:
- The redundant bridge is collapsed onto the existing x = -6.69945535 VREF rail.
- The focused test now asserts that no near-parallel VREF overlap remains.
- The solver snapshot shows the final combined output.

## Approach

Run a narrow cleanup after `NetLabelToTraceSolver` finishes, because final net-label recovery can reintroduce the redundant rail after the earlier same-net alignment stage.

A trace is collapsed only when:
- it is a complete two-point bridge;
- a same-net donor segment has matching orientation and extents within a 0.05 separation;
- both old endpoints already touch other same-net geometry;
- each old endpoint and replacement endpoint lie on the same continuous existing trace; and
- no anchored or inline label is attached to the bridge.

The trace object, ID, pins, and connectivity metadata are preserved. Only its two path points move onto geometry that already exists, so the operation does not create a new collision corridor.

## Validation

- `bun run format:check`
- `bunx tsc --noEmit`
- `bun test --timeout 30000` — 411 passed, 4 skipped, 0 failed
- Only the acoustic-tuner reproduction snapshot changed


iprasen:
Closing as superseded by #1213, which landed while this PR was being prepared. After evaluating the branch against current `main`, the new cleanup triggers zero times for the acoustic-tuner reproduction and zero times across all 173 repository JSON inputs. Keeping it would add defensive complexity without a current reproduction, contrary to the requested reproduce-before-fixing workflow.

PR #1221, opened 2026-09-21
test: reproduce ground label orientation fallback

MustafaMulla29:
## Summary
- add a minimal mouse-switch schematic trace reproduction captured from @tscircuit/core 0.0.1947 (f5cf2fab)
- preserve the exact core-generated solver input
- assert the current bug where GND is limited to y- but collision fallback relocates it to y+
- add a solver snapshot of the failing layout

## Verification
- BUN_UPDATE_SNAPSHOTS=1 bun test tests/repros/repro-mouse-switch-ground-symbol-fallback.test.ts
- bunx tsc --noEmit

This PR intentionally contains only the reproduction; the fix will be submitted separately as a stacked PR.


MustafaMulla29:
Closing in favor of #1222, which now reuses and extends the existing mouse-switch reproduction and is based directly on main.

